PR #8, opened 2016-04-03
Add travis test for gpg and gpg2

juliosueiras:


juliosueiras:
gpg 1 all pass, gpg 2  2 fail

sobolevn:
Thank you for your effort! Please, squash your commits into one, and I will merge it.
I guess the issue is with output format, because it fails inside the `uninstall_fixture_full_key` function. 
I will fix it pretty soon.

juliosueiras:
Okie, I will do that

PR #9, opened 2016-04-04
Added test for gpg and gpg2 for travis

juliosueiras:


sobolevn:
`master`-branch is protected, because `antigen` plugins rely on the `master`. So, I technically can't merge your failing PR to that branch. Could you please reopen the same PR for the `develop`?

Thanks again!

PR #16, opened 2016-05-13
Satisfy mktemp filename requirements

lrascao:
According to http://www.gnu.org/software/autogen/mktemp.html doc
the temp filename should contain at least three X characters.


sobolevn:
Thank you for your contribution. Could you please rebase your PR to the `develop` branch? 
`master` is protected. Since `antigen` is installing plugins from it.

lrascao:
sure, closing this one to switch the destination branch to develop

PR #56, opened 2016-12-08
Allow git-secret to run from any directory in the work tree

timchurchard:
Hi, this allows git-secret to work from any working directory in the git repository.  The .gitsecret directory remains in the working directory allowing users to have more than one .gitsecret in the repo.

This will result in a small change of behaviour.  If a user runs git-secret from the non-root of the repo' instead of exiting with 'not a repo' error it will exit with '.gitsecret directory not found'.

### DIFF
--- a/src/main.sh
+++ b/src/main.sh
@@ -4,7 +4,9 @@ set -e
 
 function _check_setup {
   # Checking git and secret-plugin setup:
-  if [[ ! -d ".git" ]] || [[ ! -d ".git/hooks" ]]; then
+  local is_tree
+  is_tree=$(_is_inside_git_tree)
+  if [[ $is_tree != "0" ]]; then
     _abort "repository is broken. try running 'git init' or 'git clone'."
   fi
 
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -142,6 +142,12 @@ function _add_ignored_file {
 }
 
 
+function _is_inside_git_tree {
+    git rev-parse --is-inside-work-tree >/dev/null 2>&1
+    echo $?
+}
+
+
 # Logic :
 
 function _abort {


sobolevn:
Thanks! That's a good idea. But I have some minor comments.

src/main.sh
```diff
@@ -4,9 +4,12 @@ set -e
 
 function _check_setup {
   # Checking git and secret-plugin setup:
-  if [[ ! -d ".git" ]] || [[ ! -d ".git/hooks" ]]; then
+  set +e
```
sobolevn: We use `set -e` in all our code, I guess there's no reason to break the rule here.

src/main.sh
```diff
@@ -4,9 +4,12 @@ set -e
...
+  git rev-parse --is-inside-work-tree >/dev/null 2>&1
```
sobolevn: It is better to move this function into `utils`. I guess it could be named `_is_inside_git_tree`.
And call it here like `local is_tree = $(_is_inside_git_tree)`
So, the tasks are:

- [ ] Move this function into `utils` file and `git` section
- [ ] Test it, both scenarios: inside and outside

src/main.sh
```diff
@@ -4,9 +4,12 @@ set -e
...
+  if [ $? != 0 ]; then
```
sobolevn: And here we can work with `is_tree` variable initialized above.

sobolevn:
Sorry for the failing build, I should fix the building process on OS X.

timchurchard:
Thanks for the review comments.  I've made the changes you suggested.  The _is_in_git_tree function now lives in the # VCS section in the _utils/_git_secret_tools.sh and the set +e has been removed.

sobolevn:
Done! Thanks again. 

I am planning to make a new release just around the New Year.

PR #62, opened 2017-02-11
Add -d option to hide command to remove unencrypted files.

timchurchard:
This adds an option -d to the git secret hide command to remove the unencrypted files after encryption.  I often forget to git clean -fX and leave the unencrypted files in my repo.

Tested on Ubuntu 16.10 with bash (4.3.36) & zsh (5.2) & shellcheck (0.4.4).

### DIFF
--- a/src/commands/git_secret_hide.sh
+++ b/src/commands/git_secret_hide.sh
@@ -6,7 +6,7 @@ function _optional_clean {
   local clean=0
   local opt_string=''
 
-  while getopts 'cvh' opt; do
+  while getopts 'cdvh' opt; do
     case "$opt" in
       c) clean=1;;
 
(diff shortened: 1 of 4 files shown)


src/commands/git_secret_hide.sh
```diff
@@ -27,6 +27,41 @@ function _optional_clean {
...
+        echo && echo 'removing unencrypted files:'
+      fi
+
+      while read -r line; do
+        find . -name "*$line" -type f -print0 | xargs -0 rm -f$verbose
+      done < "$SECRETS_DIR_PATHS_MAPPING"
+
+      if [[ ! -z "$verbose" ]]; then
+        echo
+     fi
```
sobolevn: One space is missing here. Indentation is broken.

tests/test_hide.bats
```diff
@@ -31,12 +31,19 @@ function teardown {
 }
 
 
-@test "run 'hide' with params" {
+@test "run 'hide' with params -c" {
```
sobolevn: Could you please rename these test to: `"run 'hide' with -c param"` and

tests/test_hide.bats
```diff
@@ -31,12 +31,19 @@ function teardown {
...
   run git secret hide -v -c
   [ "$status" -eq 0 ]
 }
 
 
+@test "run 'hide' with params -d" {
```
sobolevn: `"run 'hide' with -d param"`

sobolevn:
Thanks for your idea and implementation! I appreciate it.
Could you please fix two minor problems?

Nice work!

timchurchard:
Thanks for the suggestions, I've force pushed the fixes.

PR #64, opened 2017-02-17
Fix bug when passing more than one file to changes

msilvestre:
When passing more than one file to changes it only assumes the first one.

No

(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Any other comments?
-------------------
No

Where has this been tested?
---------------------------
**Operating system:** Mac OS 10.11.6

### DIFF
--- a/src/commands/git_secret_changes.sh
+++ b/src/commands/git_secret_changes.sh
@@ -18,7 +18,7 @@ function changes {
   shift $((OPTIND-1))
   [ "$1" = '--' ] && shift
 
-  local filenames="$1"
+  local filenames="$@"
   if [[ -z "$filenames" ]]; then
     # Checking if no filenames are passed, show diff for all files.
     filenames=$(git secret list)
--- a/src/commands/git_secret_changes.sh
+++ b/src/commands/git_secret_changes.sh
@@ -18,15 +18,15 @@ function changes {
   shift $((OPTIND-1))
   [ "$1" = '--' ] && shift
 
-  local filenames="$@"
-  if [[ -z "$filenames" ]]; then
+  local filenames=( "$@" )
+  if [[ ${#filenames[@]} -eq 0 ]]; then
     # Checking if no filenames are passed, show diff for all files.
-    filenames=$(git secret list)
+    filenames=( $(git secret list) )
   fi
 
   IFS='
   '
-  for filename in $filenames; do
+  for filename in "${filenames[@]}"; do
     local decrypted
     local content
     local diff_result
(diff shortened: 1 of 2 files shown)


sobolevn:
Thanks! That's a valuable contribution, I appreciate it.

Could you please consider some moments:

1. I have tried to run your code, it works perfectly. But one thing caught my eye. Here's the log:

> You need a passphrase to unlock the secret key for
> user: "Nikita Sobolev (keybase) <mail@sobolevn.me>"
> 2048-bit RSA key, ID 3F5E2521, created 2016-03-18 (main key ID 8AE3C73E)
> 
> changes in one.txt: 1c1,3
> < Test
> ---
> > CONFIG = 1
> > DEBUG = 1
> > TEST = 0
> 
> You need a passphrase to unlock the secret key for
> user: "Nikita Sobolev (keybase) <mail@sobolevn.me>"
> 2048-bit RSA key, ID 3F5E2521, created 2016-03-18 (main key ID 8AE3C73E)
> 
> changes in two.txt: 1c1
> < Different stuff
> ---
> > Different totelly stuff
> 

These blocks of text produces by `gpg` is breaking `diff` blocks apart. This issue is not caused by tour changes of course. But, maybe we could do something about it? I guess you have to look at `_decrypt` command. And redirect output from the `gpg` command.

2. Add a test case for this functionality. It will be quite similar to `@test "run 'changes' with multiple files changed"`

3. Fix linting:
> In src/commands/git_secret_changes.sh line 21:
>   local filenames="$@"
>                   ^-- SC2124: Assigning an array to a string! Assign as array, or use * instead of @ to concatenate.
Link: https://travis-ci.org/sobolevn/git-secret/jobs/202725065

It is fully described how to use `shellcheck` in https://github.com/sobolevn/git-secret/blob/master/CONTRIBUTING.md

4. Update docs in `man` folder.

Nice work!

msilvestre:
I am not familiar with this bash test tool. From my point of view the testa `@test "run 'changes' with multiple files changed"` is wrong, it should not be passing at all. Other proof that it is wrong is, with my change it now should be failing and it is not.

I don't have many time to work in this, but I will give it a try.

msilvestre:
Ups, I closed this by mistake.. Sorry

sobolevn:
This test seems good to me.

The logic behind it is simple: we scan for all changes in all files added to the `git-secret`.

Your logic should not break it.

msilvestre:
You are right, for some reason I though you were passing the files through the command line.

I have made a new push with shellcheck and the added test. I didn't change the man pages, I think it does not need any change, or what should be changed?

Thanks.

sobolevn:
I will merge and fix it, since I want to make a release today.

Thanks again!

PR #90, opened 2017-03-09
Make hide -d work with secrets in subdirectories

ctavan:
When searching for files to be removed after being hidden we must use
find with `-path` instead of `-name` if we want to remove files in
subdirectories as well.

–

–

Any other comments?
-------------------
–


sobolevn:
Thanks! Good one.

Issue #146, opened 2018-04-15
no_such_file test in test_cat.bats causes stall on OSX


joshrabinowitz:
What are the steps to reproduce this issue?
-------------------------------------------
* checkout current master and run 'make clean build test' on OSX

What happens?
-------------
* On OSX, may stall waiting for password at:

````
 run 'cat' with password argument
 Inserisci la passphrase:
````

joshrabinowitz:
See https://github.com/sobolevn/git-secret/pull/147/files

joshrabinowitz:
can no longer reproduce in master branch

PR #147, opened 2018-04-15
remove test that stalls sometimes

joshrabinowitz:
Prove that this is the test that causes the stall

### DIFF
--- a/tests/test_cat.bats
+++ b/tests/test_cat.bats
@@ -37,8 +37,15 @@ function teardown {
   [ "$FILE_CONTENTS" == "$output" ]
 }
 
-@test "run 'cat' with wrong filename" {
-  run git secret reveal -d "$TEST_GPG_HOMEDIR" -p "$password" NO_SUCH_FILE
-  [ "$status" -eq 2 ]
-}
+#
+# Negative test - what happens if we ask for unknown file
+#
+#  in OSX, we see this test stall running pgp under the covers with the output:
+#    ✓ run 'cat' with password argument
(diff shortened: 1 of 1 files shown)


joshrabinowitz:
See https://github.com/sobolevn/git-secret/issues/146

joshrabinowitz:
Ok, this PR confirms that it is the negative test (commented out in this PR) that causes most CI jobs to fail.

The remaining two CI failures for this branch are because of the error `"Failed to validate the SSL certificate for api.github.com:443.`

sobolevn:
The problem here is not in the test itself, but in a way that different versions of `gnupg` handles input.

It is really hard to input something in different `gnupg` versions. It differs a lot in `1.x`, `2.0`, and `2.1`.

We can try to write several different methods to do it. And to use the correct one based on version.

joshrabinowitz:
Since it seems to happen here when git secret cat tries to decrypt a file not in the .gitsecret/paths/mapping.cfg file, I'll plan to add code to only attempt to decrypt files mentioned there.

sobolevn:
So, as I understand there can be two situations:
1. We try to `cat` existing and valid secret file, test is done for this case
2. We try to `cat` file is missing from keychain, test is done, but commented. Since it is broken on some systems

Did you consider my suggestion in https://github.com/sobolevn/git-secret/pull/147#issuecomment-381431552 ? Or have you found any other reasons why this happens?

joshrabinowitz:
@sobolevn are you saying we should 'write several different methods' to handle different versions of gpg?

Issue #163, opened 2018-04-21
.ci/before_script.sh: virtualenv SNIMissingWarning and InsecurePlatformWarning


joshrabinowitz:
the script `".ci/before_script.sh"` shows the following warning:

````
[13:17:31] Ubuntu-14.04 installing some apt packages (might take some time) ... ✅ 
[13:17:33] Installing/upgrading virtualenv ...

/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:318: 
 SNIMissingWarning: An HTTPS request has been made,
 but the SNI (Subject Name Indication) extension to TLS is not available on this platform. This may 
 cause the server to present an incorrect TLS certificate, which can cause validation failures. You can
 upgrade to a newer version of Python to solve this. For more information, see 
 https://urllib3.readthedocs.io/en/latest/security.html#snimissingwarning.
  SNIMissingWarning

/usr/local/lib/python2.7/dist-packages/pip/_vendor/requests/packages/urllib3/util/ssl_.py:122: 
 InsecurePlatformWarning: A true SSLContext object is not available. This prevents urllib3 from 
 configuring SSL appropriately and may cause certain SSL connections to fail. You can upgrade to a
 newer version of Python to solve this. For more information, see 
 https://urllib3.readthedocs.io/en/latest/security.html#insecureplatformwarning.
  InsecurePlatformWarning

(InsecurePlatformWarning repeated 4 more times)
````

joshrabinowitz:
This is fixed by #303

Issue #157, opened 2018-04-18
Fix pip warnings related to incompatible pbr 5.1.1


joshrabinowitz:
We are seeing these warnings (not errors) in the CI build logs (all platforms). For example see
https://travis-ci.org/sobolevn/git-secret/jobs/368148728

````
       python-keystoneclient 3.15.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       oslo-i18n 3.20.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       oslo-utils 3.36.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       stevedore 1.28.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       oslo-serialization 2.25.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       debtcollector 1.19.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       keystoneauth1 3.4.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       python-keystoneclient 3.15.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       oslo-i18n 3.20.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       oslo-utils 3.36.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       stevedore 1.28.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       oslo-serialization 2.25.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       debtcollector 1.19.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
       keystoneauth1 3.4.0 has requirement pbr!=2.1.0,>=2.0.0, but you'll have pbr 1.10.0 which is incompatible.
[...]

PR #166, opened 2018-04-22
for #163

joshrabinowitz:
to avoid warnings about virtualenv SNIMissingWarning and InsecurePlatformWarning

### DIFF
--- a/.Dockerfiles/ubuntu/rolling/Dockerfile
+++ b/.Dockerfiles/ubuntu/rolling/Dockerfile
@@ -16,7 +16,7 @@ RUN apt-get update \
 	python2.7 \
 	sudo \
 	systemd \
- && pip install --upgrade pip==9.0.3 \
+ && pip install --upgrade pip==10.0.1 \
  && apt-get clean \
  && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* \
  && if ! getent passwd <%= @username %>; then \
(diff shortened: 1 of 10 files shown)


joshrabinowitz:
Shoot, this doesn't fix the problem:
In https://travis-ci.org/sobolevn/git-secret/jobs/369743060 (behind the 'before script' info toggle) or
https://api.travis-ci.org/v3/job/369743060/log.txt

sobolevn:
What else can we do to fix it?

joshrabinowitz:
I don't know how to fix this warning

joshrabinowitz:
See also: 
https://www.pythonanywhere.com/forums/topic/2739/

.Dockerfiles/ubuntu/rolling/Dockerfile
```diff
@@ -16,7 +16,7 @@ RUN apt-get update \
 	python2.7 \
 	sudo \
 	systemd \
- && pip install --upgrade pip \
+ && pip install --upgrade pip==10.0.1 \
```
sobolevn: Why do we have such constraint? 

It is better not to pin versions like this. Since when any updates (including security ones) will be released we won't have them here.

joshrabinowitz:
This didn't fix any of the issues I was seeking to fix.

.Dockerfiles/ubuntu/rolling/Dockerfile
joshrabinowitz: Something I read made me think this might fix the PBR issue: #157

Issue #132, opened 2018-02-22
404 http://git-secret.io/tutorials


simplenotezy:
On http://git-secret.io/ - clicking Tutorials, give a 404

sobolevn:
Yeap, that's true.
But it is a duplicate of https://github.com/sobolevn/git-secret/issues/65

joshrabinowitz:
assigning to self, at least to fix the 404. 
It looks like that html isn't in this repo. Can someone point me to where I should go to fix it?

sobolevn:
@joshrabinowitz you can create new `.md` file here: https://github.com/sobolevn/git-secret/tree/gh-pages

Just mention that tutorials are `WIP`.
Thanks!

joshrabinowitz:
Will do. Looking at https://github.com/sobolevn/git-secret/blob/master/CONTRIBUTING.md , it appears I should be starting my branch from 'develop' instead of from 'master'. My last merged branch was branched from master. 

Should I continue on the master branch, or merge master into develop and create a branch off 'develop'?

sobolevn:
We can drop `develop` branch.
It was used previously to make sure `zsh` plugins work with the stable releases only.
But we do not support them any more.

But, when working with `gh-pages` make sure that this branch is orphaned. It is not related with any other branch.

joshrabinowitz:
@sobolevn @simplenotezy I took a crack at making a guide but it wasn't easy to improve on the instructions at 
https://github.com/sobolevn/git-secret/blob/master/man/man7/git-secret.7.ronn

I propose we remove the link for "Tutorials" at the top of http://git-secret.io/git-secret
and make the link to git-secret.7.ronn more prominent. 

Thoughts?

sobolevn:
@joshrabinowitz yeap, let's do that!

joshrabinowitz:
All right!  

The `git-secret.7.fann` documentation is already included in the homepage for `git-secret`.io so no changes were needed except removal of `/tutorials.html link`.

PR #167, opened 2018-04-22
set GITSECRET_DIST once per gpg source/distribution type combination

joshrabinowitz:
As per end of https://github.com/sobolevn/git-secret/issues/132#issuecomment-381404248 ,
one dist per gpg source / DIST type combination


joshrabinowitz:
now there is just one `rpm`, `deb`, and `make` target

Issue #169, opened 2018-04-23
set -e causes silent death in _user_required() when gpg has non-fatal error


notjames:
I use two different systems for this repo. One on Darwin; the other in Linux. This problem is happening on my Linux side, though admittedly, I have had to switch to using the Darwin side because sadly, the original git-secret worked on Mac and not Linux. Now I'm trying to transition to using my Linux repo as I prefer to stay as far away from Mac as possible.

What are the steps to reproduce this issue?
-------------------------------------------
1. gpg --list-keys on my repo db gives odd but terminal error of: `gpg: skipped packet of type 12 in keybox`
2. When attempting to `git-secret reveal`, git-secret silently dies

What happens?
-------------
Nothing except `gpg: skipped packet of type 12 in keybox` multiple times to STDOUT. 

What were you expecting to happen?
----------------------------------
Some kind of error from git-secret except that the error from `gpg` above is non-fatal.

Any other comments?
-------------------
A workaround I've implemented, though not ideal, is the following lines:
```
   535   set +e                                                                                                                                                                                                      
   536   local keys_exist                                                                                                                                                                                            
   537   keys_exist=$($gpg_local -n --list-keys 2>/dev/null)                                                                                                                                                         
   538   set -e                                                                                                                                                                                                      
```

What versions of software are you using?
----------------------------------------
**Linux VM:**

**Operating system:** (`uname -a`) Ubuntu 17.10

**`git-secret` path:** (`which git-secret`) /usr/bin/git-secret

**`git-secret` version:** (`git secret --version`) 0.2.3

**`git` version:** (`git --version`) git version 2.16.2

**Shell type and version:** (`$SHELL --version`) bash 4.4.12(1)-release

**`gpg` version:** (`gpg --version`) 
gpg (GnuPG) 2.1.15
libgcrypt 1.7.8
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /home/\<me\>/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
[...]

Issue #136, opened 2018-03-16
GnuPG2 2.2 vs 2.1 conflicts in keybox format


gthank:
What are the steps to reproduce this issue?
-------------------------------------------

1. Initialize `git-secret` on a machine using `gnupg2` version 2.2
2. Use `git-secret` (`tell`, `add`, `hide`, etc.) and publish the changes
3. Clone your `git` repo onto a machine using `gnupg2` version 2.1 (w/ the proper keys imported)
4. `git secret reveal`

What happens?
-------------

```
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
gpg: skipped packet of type 12 in keybox
```

And no secrets are revealed.

What were you expecting to happen?
----------------------------------

The secrets to be revealed.

Any other comments?
-------------------

After digging around, it seems that GnuPG 2.2 added some things to the keybox format that older versions don't have. I realize this is a result of problems with underlying libraries. What I'm hoping is there's some way for you to force an interoperable format by specifying a flag when creating the keybox (or just force use of the older `.gpg` keyring format instead of the newer `.kbx`).

What versions of software are you using?
----------------------------------------

**Operating system:**

* `Darwin hank-mbpr 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64`
* `Linux falcon 4.4.0-43-Microsoft #1-Microsoft Wed Dec 31 14:42:53 PST 2014 x86_64 x86_64 x86_64 GNU/Linux`

**`git-secret` path:**

*`/usr/local/bin/git-secret`
* `/usr/bin/git-secret`

**`git-secret` version:**

* `0.2.2`
* `0.2.2`

**`git` version:**

* `git version 2.16.2`
* `git version 2.7.4`

**Shell type and version:**

* `zsh 5.4.2 (x86_64-apple-darwin17.3.0)`
[...]

PR #170, opened 2018-04-24
Detect gpg error 136

joshrabinowitz:
For #136 and #169 
Better error checking around gpg --list-keys, 
make sure we decrypt files as expected.


joshrabinowitz:
I think this PR makes things work a little better and more predictably than without.

I suggest we merge it and then consider more sophisticated messaging or runtime checks.

Thoughts @sobolevn @gthank ?

gthank:
@joshrabinowitz I think this will make it more clear what's happening, for sure. Thanks.

Issue #135, opened 2018-03-14
some commands fail when the current directory's path includes spaces


MichaelPeter-Shockoe:
Some commands, specifically at least `git secret init` and `git secret reveal`, encounter errors when the current directory's path contains any spaces. I personally avoid using spaces in directory names, but I am attempting to set up git-secret with our CI server and must work around the existing structure there.

What are the steps to reproduce this issue?
-------------------------------------------
In place of the suggested sections, I will provide two full examples which I hope will show the issue clearly.

Note: both of these examples take place on a machine on which I am able to use git-secret without issue when the path does not contain a space

### Example 1 - Initializing to a directory which contains a space
Here is the base case where I try to set up git-secret from scratch in a directory which already contains a space in the path
<img width="920" alt="screen shot 2018-03-14 at 10 17 04 am" src="https://user-images.githubusercontent.com/19493313/37409121-ed726f06-2773-11e8-9d9b-b51401d68991.png">

### Example 2 - Calling `git secret reveal` from a directory which contains a space
Here, I first initialize git-secret and encrypt a file successfully in a directory with no spaces in the path so that I can get further along in the process. I then rename the directory to contain a space and try to call `git secret reveal`
<img width="927" alt="screen shot 2018-03-14 at 10 19 47 am" src="https://user-images.githubusercontent.com/19493313/37409298-5aaa0624-2774-11e8-9657-976dd8b61516.png">

Any other comments?
-------------------
I inserted some logging and played with the code a bit, and I believe the issue is stemming from lines of the form
```bash
$($gpg_local ......)
```

A little research pointed me towards the use of `eval` for better interpretation of commands formed by concatenating strings, but I am not sure of the security implications around that command.

What versions of software are you using?
----------------------------------------
**Operating system:** macOS 10.13.3
```
Darwin -- 17.4.0 Darwin Kernel Version 17.4.0: Sun Dec 17 09:19:54 PST 2017; root:xnu-4570.41.2~1/RELEASE_X86_64 x86_64
```

**`git-secret` path:** /usr/local/bin/git-secret

**`git-secret` version:** 0.2.2

**`git` version:** 2.11.1

**Shell type and version:** zsh 5.3 (x86_64-apple-darwin17.0)

**`gpg` version:** 2.2.5

sobolevn:
@MichaelPeter-Shockoe Thanks for the detailed report.

Are you interested in providing a fix?

MichaelPeter-Shockoe:
[...]

PR #175, opened 2018-05-10
code simplification

joshrabinowitz:


src/_utils/_git_secret_tools.sh
```diff
@@ -213,7 +213,8 @@ function _gawk_inplace {
 
   _temporary_file
 
-  bash -c "gawk ${parms}" > "$filename"
+  # below doesn't handle directories with spaces correctly
```
sobolevn: As my quick research showed there's also an issue described in #135 in this line: https://github.com/sobolevn/git-secret/blob/deae0d1cd53d9cb85f3d34992a1afef866842a27/src/_utils/_git_secret_tools.sh#L617

That's why this error happens:

```
path /Users/sobolev/Desktop/vue_docker/folder with spaces/file.txt
gpg: Note: '--pinentry-mode' is not considered an option
gpg: Note: '--quiet' is not considered an option
usage: gpg [options] --decrypt [filename]
```

src/_utils/_git_secret_tools_linux.sh
```diff
@@ -8,8 +8,7 @@ function __replace_in_file_linux {
 
 
 function __temp_file_linux {
-  local filename
-  filename=$(mktemp)
+  local filename=$(mktemp)
```
sobolevn: I guess this will break `shellcheck`. That was a weird rule.
I can not fully explain why it is required.

src/_utils/_git_secret_tools.sh
```diff
@@ -630,7 +630,7 @@ function _decrypt {
   fi
 
   if [[ ! -z "$homedir" ]]; then
-    base="$base --homedir=$homedir"
+    base="$base --homedir \"$homedir\""
```
sobolevn: Does it solve the whitespace issue? 
I have also tried different approaches with no luck.
But I did not try this one, I guess.

I have even tried to rewrite the whole function using array of arguments.

joshrabinowitz: I don't think it does solve the issue

joshrabinowitz: but yes, using arrays seems to be the answer

src/_utils/_git_secret_tools.sh
joshrabinowitz: I'm not sure that's right. I think the error happens because we're trying to pass the parameters as a string instead of as an array.

Issue #179, opened 2018-05-14
travis-ci can't find man pages during tests on ubuntu-rolling and ubuntu-latest


joshrabinowitz:
See 
https://travis-ci.org/sobolevn/git-secret/jobs/378803909 
and
https://travis-ci.org/sobolevn/git-secret/jobs/378803908

````
       Use `bundle info [gemname]` to see where a bundled gem is installed.
       Running Serverspec
       ---> RSPEC_CMD variable is: /usr/local/bin/rspec
       
       git-secret::test
         Package "git-secret"
           should be installed
         Command "find /tmp/git-secret/build -name "*.deb""
           stdout
             should match /git-secret.*deb/
         File "/.git-secret_test-passed"
           should exist
         File "/.git-secret_lint-passed"
           should exist
         Command "dpkg-query --status git-secret"
           exit_status
             should eq 0
         Command "man -w "git-secret""
           exit_status
             should eq 0 (FAILED - 1)
         Command "man -w "git-secret-init""
           exit_status
             should eq 0 (FAILED - 2)
       
       Failures:
       
         1) git-secret::test Command "man -w "git-secret"" exit_status should eq 0
            Failure/Error: its(:exit_status) { should eq 0 }
              
              expected: 0
                   got: 16
              
              (compared using ==)
              env PATH="/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin" /bin/sh -c man\ -w\ \"git-secret\"
              
            # ./roles/git-secret/.ci/integration/gnupg1/serverspec/default_spec.rb:46:in `block (3 levels) in <top (required)>'
       
         2) git-secret::test Command "man -w "git-secret-init"" exit_status should eq 0
            Failure/Error: its(:exit_status) { should eq 0 }
              
              expected: 0
                   got: 16
              
              (compared using ==)
              env PATH="/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin" /bin/sh -c man\ -w\ \"git-secret-init\"
              
            # ./roles/git-secret/.ci/integration/gnupg1/serverspec/default_spec.rb:50:in `block (3 levels) in <top (required)>'
       
       Finished in 0.09844 seconds (files took 0.41229 seconds to load)
       7 examples, 2 failures
       
       Failed examples:
       
[...]

PR #180, opened 2018-05-15
no change

joshrabinowitz:
testing current build


joshrabinowitz:
@sobolevn there are some new, env-specific build failures appearing in the latest ubuntu, debian and fedora test runs. 
Stuff related to man pages not being found; see #179

PR #181, opened 2018-05-17
migrate from bats to bats-core

joshrabinowitz:
bats is no longer maintained

PR #193, opened 2018-06-09
migrate from bats to bats-core 

joshrabinowitz:
bats is no longer maintained.
This PR replaces #181


joshrabinowitz:
I say we merge this before marking 0.2.4 as released. @sobolevn @simbo1905 ?

Issue #201, opened 2018-06-12
Manual installation fails


eburgueno:
What are the steps to reproduce this issue?
-------------------------------------------
```
[user@host ~]$ git clone https://github.com/sobolevn/git-secret.git
Cloning into 'git-secret'...
remote: Counting objects: 3132, done.
remote: Compressing objects: 100% (7/7), done.
remote: Total 3132 (delta 2), reused 7 (delta 2), pack-reused 3123
Receiving objects: 100% (3132/3132), 2.14 MiB | 2.08 MiB/s, done.
Resolving deltas: 100% (1983/1983), done.
[user@host ~]$ cd git-secret/
[user@host git-secret]$ make build
[user@host git-secret]$ PREFIX="~/opt/git-secret-0.2.4" make install
cp: cannot stat ‘/home/user/~/git-secret/utils/bin/*’: No such file or directory
make: *** [install] Error 1
```

What happens?
-------------
Error as above

What were you expecting to happen?
----------------------------------
git-secret is installed in $PREFIX

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`)
Linux host.example.com 3.10.0-693.21.1.el7.x86_64 #1 SMP Wed Mar 7 19:03:37 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

CentOS Linux release 7.4.1708 (Core) 

**`git` version:** (`git --version`)
git version 2.14.1

**Shell type and version:** (`$SHELL --version`)
GNU bash, version 4.2.46(2)-release (x86_64-redhat-linux-gnu)

**`gpg` version:** (`gpg --version`)
gpg (GnuPG) 2.0.22
libgcrypt 1.5.3

ilovezfs:
Seeing the same thing in Homebrew CI for 0.2.4 CC @sobolevn

joshrabinowitz:
I think the recently-merged #203 fixes this

sobolevn:
@ilovezfs yeap, since `brew` uses the same install script as manual installation.

eburgueno:
#203 seems to have fixed it. Thank you.

```
[user@host ~]$ git clone https://github.com/sobolevn/git-secret.git
Cloning into 'git-secret'...
remote: Counting objects: 3135, done.
remote: Compressing objects: 100% (9/9), done.
remote: Total 3135 (delta 3), reused 10 (delta 3), pack-reused 3123
[...]

PR #203, opened 2018-06-12
Revert "migrate from bats to bats-core "

joshrabinowitz:
Reverts sobolevn/git-secret#193
I think this causes #201, 'manual installation fails'
(The problem is in  utils/install.sh)

I'll recreate the port from bats to bats-core without this issue at a later time.

### DIFF
--- a/.ci/integration/tasks/dependencies.yml
+++ b/.ci/integration/tasks/dependencies.yml
@@ -14,7 +14,7 @@
 
 - name: Get bats
   git:
-    repo: 'https://github.com/bats-core/bats-core.git'
+    repo: 'https://github.com/sstephenson/bats.git'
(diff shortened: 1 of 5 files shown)


joshrabinowitz:
Before PR:
````
% PREFIX=/tmp/git-secret make install
cp: /Users/joshr/gitsrc/git-secret/utils/bin/*: No such file or directory
````

After PR: 
````
% PREFIX=/tmp/git-secret make install
Installed git-secret to /tmp/git-secret/bin/git-secret
````

jasonkarns:
Is there an issue that needs opened on bats-core related to this?

joshrabinowitz:
@jasonkarns no, no issue is needed on bats-core.

Issue #202, opened 2018-06-12
Silent error on secret hide when GPG key expires


rjcoelho:
When a GPG key expires we get no output when doing a "git secret hide". We should get a GPG key expired error message (if possible).

sobolevn:
Yeap, that's true. This check should also happens when we try to add (`git secret tell`) an expired key.

rjcoelho:
Basically any call to gpg should check for error codes and report then to stderr (and stop the command). Again if this is possible.

joshrabinowitz:
Could someone write a test in a PR to show the problem?

joshrabinowitz:
@rjcoelho I think this is what you're asking for: PR #205 'more careful error checking running gpg'

joshrabinowitz:
I think we can close this, @rjcoelho please let us know if PR #205  fixes the issues you raise here.

rjcoelho:
@joshrabinowitz it solve my issue. Thanks for such a speedy solution.

joshrabinowitz:
re-opened until #205 is in the released code.

joshrabinowitz:
this is a dup o #238

PR #205, opened 2018-06-12
more careful error checking running gpg

joshrabinowitz:
for #202, " Silence error on secret hide when GPG key expires"

Checks return codes of invocations of gpg

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -553,7 +553,7 @@ function _user_required {
     # this might catch corner case where gpg --list-keys shows 
     # 'gpg: skipped packet of type 12 in keybox' warnings but succeeds? 
     # See #136
-    _abort "problem listing public keys in gpg: exit code $exit_code"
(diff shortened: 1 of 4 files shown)


joshrabinowitz:
@sobolevn @simbo1905 you cool with me merging?

joshrabinowitz:
@simbo1905 is this PR in the 0.2.4 release?

simbo1905:
Until then - no fix merges. Can some else take a look at making travis only deploy from tags as I need to focus on some other things at the moment.

joshrabinowitz:
* Well the good new is that all the versions of 0.2.4 should interoperate well (newer PRs have only changes for bats->bats-core, error messaging, and docs.)

* I agree, releases should not happen on 'merge to master' but instead when we tag the code.

* Yes, let's do a 2.5.0 tag/release when convenient.

joshrabinowitz:
Also we should re-open #202 until the fix is released. Doing so now.

PR #206, opened 2018-06-13
migrate from bats to bats-core (redux)

joshrabinowitz:
Redo of PR #193 without changes to utils/install.sh


joshrabinowitz:
went ahead and merged this because it's the same code changes as in the approved PR #193

PR #207, opened 2018-06-14
RFC 0001 A stable and forwards compatible public key storage format

simbo1905:
This is an RFC for review, discussion, shared design.


simbo1905:
I took the format from https://gist.github.com/michaelcurry/e0132058fcd6d588a1299afd69638df4

I have only put in the first part of the discussion from https://github.com/sobolevn/git-secret/issues/136 as want to see how folks feel about the approach to this and the format (in addition to the actual design!)

I am hoping that reviewers can edit this PR and will rather than the entire burden is on the initial author to get it into the final state. Let us see how it works out...

simbo1905:
Apologies all that this PR has been so noisy. I should have perfected it on a fork but I was keen to have an early draft looked. When the build finally passes I will merge.

Issue #209, opened 2018-06-19
bats build issues on master branch


simbo1905:
A rebuild of the last successful Travis master build (I cancelled the rpm and deb deploy jobs) gives broken jobs use to a bats install error:

```TASK [Install bats] ************************************************************
       fatal: [localhost]: FAILED! => {"changed": false, "msg": "src file does not exist, use \"force=yes\" if you really want to create the link: /opt/bats/libexec/bats", "path": "/usr/bin/bats", "src": "/opt/bats/libexec/bats", "state": "absent"}
       	to retry, use: --limit @/tmp/kitchen/default.retry
```

See https://travis-ci.org/sobolevn/git-secret

This is blocking testing any improvements to the build/release/deploy system and probably any future PR.

joshrabinowitz:
I presume it's an upstream change in bats-core, this test passed 5 days ago. Will investigate

joshrabinowitz:
I think this is the change in bats-core that caused the build breakage.
https://github.com/bats-core/bats-core/commit/e8a2774534c070990b4b13818ac8ddacae6009c5

PR #210, opened 2018-06-19
Revert "migrate from bats to bats-core (redux)"

joshrabinowitz:
Reverts sobolevn/git-secret#206
For #209


joshrabinowitz:
Upstream changes in the master branch of bats-core were causing issues in the git-secret build. This reverts the change to use 'bats' instead of 'bats core'.

In the medium-term a migration to bats-core is probably a good idea. See releated issue: 'feature: migrate from bats to bats-core #211'

Issue #197, opened 2018-06-10
deploy from a git tag


simbo1905:
There was a note in #120 ”We need to run deploy on tags, right now we are using deploys on master branch”. This ticket is to track that build enhancement.

sobolevn:
This should be a really easy feature.

simbo1905:
Approach:
1) fixes and/or features are merged to master 
2) changelog is generated and manually updated as necessary 
3) contributor applies a release tag vX.Y.Z
4) travis build runs for the new tag and the travis deploy steps check that the coding being built is a release tag. if so they run.

If there are transient errors some but not all release artifacts will deploy. We then need to kick-off the build again. Best we automate that. But we now face a choice:

1) contributor re-applies a release tag vX.Y.Z 
2a) deploy steps run as before and overwrite any previous release 
2b) deploy steps look for an existing release and refuse to run assuming it's a rebuild to fix another deploy step 

Assuming we are prompt and tidy there isn't much risk of overwrite release. It would make me nervous doing that with a production service as bad things happen when the principal of immutable releases are relaxed.

What are peoples thoughts on 2a vs. 2b?

sobolevn:
I usually reapply tags manually when something goes wrong on a release.
But, I do it for cases when I am 100% sure that release was not already published.

simbo1905:
@joshrabinowitz do you have a preference?

joshrabinowitz:
For this release especially, I think @sobolevn's suggestion is good. Let's get this out the door expediently. 
In the long run I agree releases should be immutable.

joshrabinowitz:
@simbo1905 @sobolevn should we close this? I see #212 was merged.

PR #212, opened 2018-06-19
Deploy from release tag

simbo1905:

The deploy logic runs on tags with a name matching `/^v.*$/`

#197 deploy from a git tag

### DIFF
--- a/.ci/integration/tasks/dependencies.yml
+++ b/.ci/integration/tasks/dependencies.yml
@@ -14,7 +14,7 @@
 
 - name: Get bats
   git:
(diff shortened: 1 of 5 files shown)


sobolevn:
Just one notice: why `all_branches` over `branch: master`?

simbo1905:
@sobolevn if there was any `branch:` entry it would output the following for any matching tag:

`Skipping a deployment with the bintray provider because this branch is not permitted: v0.0.4`

Trying `all_branches: true` was my 7th attempt. As the deploy will only runs when the tag matches `v.*` it should be immaterial.

Issue #195, opened 2018-06-09
Automate the homebrew release for MacOS


simbo1905:
Publishing releases to `brew` is a manual process. This could be automated like the publication of the other releases.

simbo1905:
Needs #194

PR #214, opened 2018-06-19
travis build on osx

simbo1905:

Has travis run the tests on osx. 

No, but it is work towards #195 to automate the brew release.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -79,10 +79,11 @@ matrix:
       sudo: required
       language: ruby
       rvm: 2.4
-#    - os: osx
-#      env: GITSECRET_DIST="brew"
-#      sudo: false
-#      language: generic
+    - os: osx
+      env: GITSECRET_DIST="brew"
+      sudo: required
+      language: ruby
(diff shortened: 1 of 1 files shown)


simbo1905:
Given that bats-core is new testing on osx is a good idea.

simbo1905:
> @joshrabinowitz commented on this pull request.
> ...
> I see some unexpected errors in the build output that aren't in the other build though, they irk me a little
> ... 
> Can we/do we care to fix those errors?

Looking at the travis.yaml its short and sweet about osx and doesn't have `yarn` or  `shell_session_update` so I can only assume that it is travis osx plumbing. 

The `shell_session_update` is complained about here https://github.com/travis-ci/travis-ci/issues/6307

The `yarn` error is complained about here https://github.com/travis-ci/travis-ci/issues/9454

I think for the time being we simply ignore them as trying to mess around with picking osx images and installing custom packages to get around none failing error in the travis osx job runners that could be patched at any time doesn't seem like a good use of time.

PR #217, opened 2018-06-21
optimize git checkout of v1.0.2 tag of bats-core

joshrabinowitz:
As per @simbo1905 's comment on PR #216:
>can we do --depth=1 so that it takes a very shallow copy of the code rather 
>than a full deep history which will slow things down over time?


simbo1905:
LGTM

Issue #221, opened 2018-06-28
git secret + commit hook fails with git gui. Overrides pinentry


ieugen:
What are the steps to reproduce this issue?
-------------------------------------------
1. Add a pre commit hook as per documentation:

.git/hooks/pre-commit
```
 git secret changes | grep "+++"
 if [ $? -eq 0 ]; then
     echo "Hiding git secretes."
     git secret hide
 else
     echo "No secrets to hide."
 fi
```
I'm hiding secrets only if there are changes.
2. Open git gui and commit https://git-scm.com/docs/git-gui 

What happens?
-------------

Git gui hangs - because the gpg pass phrase is presented on standard input and I can't access that.

I believe I've tracked the code to this part in `/usr/bin/git-secret`
````
if [[ "$GPG_VER_21" -eq 1 ]]; then
    base="$base --pinentry-mode loopback"
fi
````

What were you expecting to happen?
----------------------------------

I should be presented with the gui pinentry application that I have setup in my `/.gnupg/gpg-agent.conf` via `pinentry-program /usr/bin/pinentry-gnome3` . 

Any other comments?
-------------------

I think Arch docs are explain gpg situation pretty good: https://wiki.archlinux.org/index.php/GnuPG#gpg-agent

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`) Linux default 4.16.0-2-amd64 #1 SMP Debian 4.16.12-1 (2018-05-27) x86_64 GNU/Linux

I'm using Debian testing (Buster). 

**`git-secret` path:** (`which git-secret`) /usr/bin/git-secret
**`git-secret` version:** (`git secret --version`) 0.2.2
**`git` version:** (`git --version`) 2.18.0
**Shell type and version:** (`$SHELL --version`) zsh  5.5.1 

**`gpg` version:** (`gpg --version`) gpg (GnuPG) 2.2.8
libgcrypt 1.8.3

joshrabinowitz:
@ieugen thank you for this report.

Can you please try again with the most recent release of git-secret, 0.2.4?

Also, if you know what code change is required to fix this issue, please let us know or open a PR.

ieugen:
Thanks. I will give it a try. Unfortunatelly I don't know the fix but I will give it a try. Gpg is not the friendliest of softwares.

joshrabinowitz:
Hello, @ieugen, @simbo1905 , @sobolevn :

What's the right way to handle this?  

[...]

PR #223, opened 2018-06-28
Fixes for GPG pinentry issue with git gui

ieugen:
Closes https://github.com/sobolevn/git-secret/issues/221

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -632,7 +632,7 @@ function _decrypt {
     base="$base --homedir=$homedir"
   fi
 
-  if [[ "$GPG_VER_21" -eq 1 ]]; then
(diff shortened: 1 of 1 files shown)


joshrabinowitz:
Thanks for the patch, @ieugen ! Does this fix your --pinentry issue? 

Unfortunately it breaks the git-secret self-tests. 
We need to find the right condition under which to avoid the` --pinentry-mode loopback` option

joshrabinowitz:
Hey @ieugen, ok with you if I close this PR? Issue #221 still tracks the problem.

joshrabinowitz:
Hello @ieugen , I think I've worked up a solution for this issue. 

Could you possibly test this case along with the code from PR #490 
with the env var `SECRETS_PINENTRY` set to `default` and then `ask`?
I think this will allow `gpg-agent` to use its normal pinentry method.

PR #224, opened 2018-06-30
allow spaces in path

joshrabinowitz:
Fixes issue #135, regarding spaces in pathnames.

Works around issue with `_get_gpg_local()` function when returning pathnames with spaces.

We could store the results of `_get_secrets_dir_keys()` in a global variable if that's preferable. Reviewers, let me know.


joshrabinowitz:
@sobolevn are you ok with this being merged?

PR #240, opened 2018-07-15
fix for keys with only emails, and no usernames. Closes #227

### LINKED ISSUE #227: keys with no info but the email address not recognized by whoknows
What are the steps to reproduce this issue?
-------------------------------------------
1. have a key in the keyring with no info but the email address: gpg will show the email address without the angle brackets; let's say john.smith@example.com
2. do `git secret tell john.smith@example.com`: the key will be correctly added to the keyring
3. do `git secret whoknows`: the key just added with `tell` won't appear anywhere and won't be considered for encryption

What happens?
-------------
See above. The problem seems to be due to the regexps in the script assuming the angle brackets around the email address, which is not the case if the key is generated with `gpg --quick-generate-key`, for example.

What were you expecting to happen?
----------------------------------
I expected the key to be revealed by `whoknows` and used

Any logs, error output, etc?
----------------------------
Not really, but it's easy to reproduce

Any other comments?
-------------------
None

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`) 
Linux nelson 4.9.0-6-amd64 #1 SMP Debian 4.9.88-1+deb9u1 (2018-05-07) x86_64 GNU/Linux

**`git-secret` path:** (`which git-secret`) 
/usr/bin/git-secret

**`git-secret` version:** (`git secret --version`) 
0.2.4

**`git` version:** (`git --version`) 
git version 2.11.0

**Shell type and version:** (`$SHELL --version`) 
GNU bash, version 4.4.12(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2016 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

**`gpg` version:** (`gpg --version`) 
gpg (GnuPG) 2.1.18
libgcrypt 1.7.6-beta
Copyright (C) 2017 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /home/bronto/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2

joshrabinowitz:
Closes #227.  
* Doesn't assume emails are surrounded in `<>`, 
* uses `cut -d: -f10` to fetch 10th field from `gpg --export-key --with-colon`

* Note: this is the first use of 'cut' in git-secret.


README.md
```diff
@@ -37,6 +37,7 @@ See the [installation section](http://git-secret.io/installation) for the detail
 - `git` since `1.8.3.1`
 - `gpg` since `gnupg 1.4` to `gnupg 2.X`
 - `sha256sum` since `8.21`
+- `cut` (any version)
```
sobolevn: Isn't `cut` included in POSIX?

joshrabinowitz: I thought POSIX defined just how commands were supposed to work and their interfaces? 

Any any case, I think we are pretty much guaranteed that `cut` will be installed  -- it's part of GNU's `coreutils`.

Here's one discussion of the topic https://unix.stackexchange.com/questions/37064/which-are-the-standard-commands-available-in-every-linux-based-distribution

simbo1905: We already have gawk and that does printing columns so do we really need cut? 

$ echo "1:2:3:4:5:6:7:8:9:10:11" | cut -d: -f10
10
$ echo "1:2:3:4:5:6:7:8:9:10:11" | awk -F':' '{print $10}'
10

simbo1905: however that's a minor matter so in general LTGM

joshrabinowitz: I feel like if we can move away from awk scripts for extremely simple tasks (in general) we win on readability and maintainability.

It's a particularly simple use of awk here though, so I agree, let's use awk here too.

Issue #172, opened 2018-04-26
Apply Unix permissions from `.secret` file to revealed file


gthank:
I noticed today that revealed files don't get the execute bit from the `.secret` version. I can work around this during my deploys by remembering to set the permissions explicitly, but it would be nice if `git-secret` automatically chmodded the revealed file to have the same perms as the encrypted original.

EDIT: On *nix machines, it should be essentially a one-liner:

```
chmod `stat -f "%Op" $src_file` $target_file
```

after the [`_decrypt` call](https://github.com/sobolevn/git-secret/blob/master/src/commands/git_secret_reveal.sh#L43) in the `reveal` command.

It might also make sense to do the same thing in the other direction when hiding files.

The trick is whether or not this would work on Windows devices (using WSL), and what to do in non-WSL environments (if those are in fact supported).

sobolevn:
@gthank nice one! do you want to elaborate on this?

joshrabinowitz:
Maybe this would make sense to be activated with a `-p` option, like `tar` and `cp` have.

gthank:
@sobolevn My use-case is that I am hiding a shell script that has some credentials in it. As it currently stands, I have to go back and chmod that plaintext file after I reveal it. I think it would be more convenient if `git-secret` did that for me. After all, `git` is already tracking the mode on the `.secret` file, so `git-secret` "knows" (or more realistically, can find out via the `stat` command) what we want the mode to be on the plaintext version. I'm like 99% sure the one-liner I proposed would also pick up things like the sticky bit and setuid/setguid, which I think would be the desired behavior.

If you mean "elaborate on how that actually works", `stat` pulls basically everything the OS knows about the file, and then the `-f "%Op"` part says "show me the permissions (`p`) and format it in octal (`%O`).

@joshrabinowitz For `cp`, the `-p` flag lets you pass a bunch of fine-grained controls over whether to preserve the mode, the timestamp, xattrs, etc., and the default is to preserve the mode (along w/ ownership and timestamps). I don't really use `tar` often enough to say for sure whether it has the same defaults, but I would guess it does. I guess it might make `git-secret` a better command-line citizen if it accepted the same basic args as `cp` or `tar`, but it seems like a lot more work (for one thing, I'm not sure how you'd "preserve" the timestamps, since it's not an actual copy operation; you'd need stronger shell-fu than mine for that part), and I'm not sure how often people would want to do it.

gthank:
[...]

PR #245, opened 2018-07-22
add -P (preserve permission) option to reveal and hide. For #172

joshrabinowitz:
For #172. This adds a -P option to 'git secret reveal' like cp's -p (preserve) option.

(-p is already used for 'password' by git-secret)

Suggestions on code changes or specific tests to add welcome.

@sobolevn @gthank  @simbo1905

### DIFF
--- a/src/_utils/_git_secret_tools_linux.sh
+++ b/src/_utils/_git_secret_tools_linux.sh
@@ -21,6 +21,6 @@ function __get_octal_perms_linux {
   local filename
   filename=$1
   local perms
-  perms=$(stat --format "'%a'" "$filename")
+  perms=$(stat --format '%a' "$filename")
   echo "$perms"
 }
--- a/src/_utils/_git_secret_tools_linux.sh
+++ b/src/_utils/_git_secret_tools_linux.sh
@@ -16,3 +16,9 @@ function __temp_file_linux {
 function __sha256_linux {
   sha256sum "$1"
 }
+
+function __get_octal_perms_linux {
+   local file=$1
+   perms=$(stat --format "'%a'" "$filename")
+   echo "$perms"
+}
--- a/src/_utils/_git_secret_tools_linux.sh
+++ b/src/_utils/_git_secret_tools_linux.sh
@@ -18,7 +18,9 @@ function __sha256_linux {
 }
 
 function __get_octal_perms_linux {
-   local file=$1
-   perms=$(stat --format "'%a'" "$filename")
-   echo "$perms"
+  local filename
+  filename=$1
+  local perms
+  perms=$(stat --format "'%a'" "$filename")
+  echo "$perms"
 }
--- a/src/_utils/_git_secret_tools_osx.sh
+++ b/src/_utils/_git_secret_tools_osx.sh
@@ -22,6 +22,6 @@ function __get_octal_perms_osx {
   local filename
   filename=$1
   local perms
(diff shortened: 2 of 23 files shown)


gthank:
This looks awesome. Thanks a ton! I still wish it was the default, but just having the option is enough to save me a ton of work, so really: thanks a lot for this work.

joshrabinowitz:
@gthank you're welcome! Yeah, we can't change default behavior until @sobolevn desires it and we reach 0.3.0.

I'm not super-happy using -c instead of -p, suggestions welcome. 

Also, if `git-secret reveal -c` cannot set the permissions of the decrypted file, and should it be a warning, or an error?

joshrabinowitz:
Also, should there be a similar setting on 'git-secret hide' to set the perms of the hidden file to match the perms of the uncompressed one?

gthank:
I'm inclined to say it should be an error. If people are relying on it to set the permissions, and then it doesn't, things are likely to break later, and be harder to spot.

As for `-c` vs `-p`, I'm not sure. If the convention is to always have a "short" version, then `-p` is already taken, and "preserve permissions" doesn't really offer an obvious alternative. `-c` for `--chmod` is as good a mnemonic as any I can think of at the moment. If you want to only provide the "long" form, then I'd be fine with using `--preserve`.

gthank:
And yes, a parallel flag on either `add` or `hide` to preserve the permissions would be ideal.

joshrabinowitz:
Re: a `--preserve` option: at this point all of git-secret uses short (single-letter) options.

joshrabinowitz:
I should get around to finishing this next week.

src/_utils/_git_secret_tools_linux.sh
```diff
@@ -16,3 +16,11 @@ function __temp_file_linux {
 function __sha256_linux {
   sha256sum "$1"
 }
+
```
joshrabinowitz: @sobolevn @simbo1905 

I'm welcoming input on the way to implement the -c option and get the 'permission string' to use for chmod? 

* I figured octal was the best representation to use. 
Agreed?

* I couldn't find just one incantation that would return the octal permissions on OSX and linux, so I'm using `stat --format '%a'` on linux and `stat -f '%p'` on osx. 
Let me know if you see a better way.

src/_utils/_git_secret_tools_linux.sh
```diff
@@ -16,3 +16,11 @@ function __temp_file_linux {
...
+function __get_octal_perms_linux {
+  local filename
+  filename=$1
+  local perms
+  perms=$(stat --format '%a' "$filename")
```
joshrabinowitz: Here we use `stat --format '%a'`

src/_utils/_git_secret_tools_osx.sh
```diff
@@ -18,3 +18,10 @@ function __temp_file_osx {
 function __sha256_osx {
   /usr/bin/shasum -a256 "$1"
 }
+function __get_octal_perms_osx {
+  local filename
+  filename=$1
+  local perms
+  perms=$(stat -f '%p' "$filename")
```
joshrabinowitz: Here we use `stat -f '%p'`

joshrabinowitz:
I think this is ready for review and/or merging, @gthank @simbo1905 @sobolevn

Issue #232, opened 2018-07-12
build issue on alpine-latest: tasks 'Create git-secret apk package' and 'Create git-secret apk location'


joshrabinowitz:
builds are failing only on alpine-latest because of error(s) with 
[run-tests.yml](.ci/integration/tasks/run-tests.yml)

For exact error messages, please see the 'alpine-latest' jobs here:
https://travis-ci.org/sobolevn/git-secret/builds/403141496
or below

joshrabinowitz:
Error message:

TASK [Create git-secret apk package] *******************************************

[...]

PR #246, opened 2018-07-22
Build fixes for Alpine latest

simbo1905:

Compile and package issues on alpine latest. 

#232 build issue on alpine-latest: tasks 'Create git-secret apk package' and 'Create git-secret apk location'

### DIFF
--- a/.ci/integration/gnupg-git/default.yml
+++ b/.ci/integration/gnupg-git/default.yml
@@ -65,12 +65,6 @@
       when:
         - ansible_os_family == "Alpine"
 
-    # here we ignore a none zero status code as the package is installed
-    - name: Install apk packaging dependencies for Alpine based distros
-      command: bash -lc "gem install etc || :"
-      when:
-        - ansible_os_family == "Alpine"
-
     - name: Install rspec in /usr/local/bin for RedHat based distros
       command: bash -lc "gem install -n /usr/local/bin rspec"
       when:
--- a/.ci/integration/gnupg-git/default.yml
+++ b/.ci/integration/gnupg-git/default.yml
@@ -90,6 +90,11 @@
       retries: 5
       delay: 10
 
+    - name: Make directory /usr/local/src/ for Alpine based distros
+      command: bash -lc "mkdir -p /usr/local/src/"
+      when:
+        - ansible_os_family == "Alpine"
+
     - name: Extract gnupg source tarball
       unarchive:
         src: /tmp/gnupg.tar.gz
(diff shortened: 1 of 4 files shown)


.ci/integration/tasks/dependencies.yml
```diff
@@ -28,3 +28,9 @@
     name: fpm
     state: present
     user_install: no
+
+# here we ignore a none zero status code as the package is installed
```
joshrabinowitz: non-zero status? (typo)

.ci/integration/gnupg-git/default.yml
```diff
@@ -132,6 +142,15 @@
       command: bash -lc "cd {{ gpg_src_path }} && ./configure {{ gpg_build_config }}"
       changed_when: False
 
+    # gnupg on alpine is not detecting absence of fig2dev when this thread suggests it should: https://lists.gnupg.org/pipermail/gcrypt-devel/2013-December/002710.html
+    - name: Disable making docs on Alpine
```
joshrabinowitz: What is the consequence of not making docs on Alpine? 
[It seems like we still generate man pages for the apk?]

.ci/integration/tasks/dependencies.yml
```diff
@@ -28,3 +28,9 @@
...
+- name: Install apk packaging dependencies for Alpine based distros
+  command: bash -lc "gem install etc || :"
```
joshrabinowitz: why the `|| :` ? Are we expecting `gem install etc` to fail sometimes?

.ci/integration/gnupg-git/default.yml
simbo1905: Good question that needs clarification in the comments. These steps hit github and pull the GPG code and builds that which wants ImageMagik to conversion of SVG to PND and other graphical stuff that git-secret doesn't use it it's man page docs. The fact that we don't generate the graphics for GPG docs has no effect on all our git-secret tests, packaging and docs.

.ci/integration/tasks/dependencies.yml
simbo1905: We are expecting it to succeed in installing `etc` then return a non-zero status code. That confused me for a while until I logged onto the Alpine docker image and confirmed that is what it does. It says it installed `etc` then barfs a complaint to stderr about some unrelated. Since it said it installed the missing package and the code that uses dosen't blow up with a missing package I chose to not to attempt attempt fix the gem system. The fix/build/test cycle is very slow to debug and wipe images to restart. This means that cleaning up ”fugly” things like this which annoy me feels excessively costly.

simbo1905: i have changed this to read `# here the package is installed but a non-zero status code is returned that we ignore`

simbo1905:
Okay, I have made changes based on Josh's feedback. 

If everyone is okay with re-enabling Alpine builds I will "squash and merge" this PR when the next build completes.

PR #247, opened 2018-08-14
Allow per-command configuration of custom secrets dir

lrascao:
Allow per-command configuration of custom secrets dir

Typical use case is when you want different people to have access
to some files and other people access to other different files.

eg.
`SECRETS_DIR=.gitsecret2 git secret init`
`SECRETS_DIR=.gitsecret2 git secret hide`

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Folders:
-_SECRETS_DIR=".gitsecret"
+_SECRETS_DIR=${SECRETS_DIR:-".gitsecret"}
 _SECRETS_DIR_KEYS="${_SECRETS_DIR}/keys"
 _SECRETS_DIR_PATHS="${_SECRETS_DIR}/paths"
 


lrascao:
sure, in my project i have several secret files that all the team has access to, those files are split by environment (live, dev, staging, etc), i need to give access a different set of files with secrets to some other external team so i can't just add those users to my team's `.gitsecret`, it's a simple way of segmenting users and secrets

joshrabinowitz:
Hm, seems like using a separate git-secret repo for each environment (live, dev, stagin) would cause less confusion? 

How would a user keep track of which files are handles by which .gitsecret directory?

lrascao:
> Hm, seems like using a separate git-secret repo for each environment (live, dev, stagin) would cause less confusion?
It might, if we had designed it that way from the ground up, not sure all of the projects using git-secret might have made the same decision as we did

> How would a user keep track of which files are handles by which .gitsecret directory?
In our case a Jenkins build with some logic around it would take care of that

joshrabinowitz:
> In our case a Jenkins build with some logic around it would take care of that

I was hoping for something not external to git-secret

lrascao:
@joshrabinowitz what do you have in mind?

joshrabinowitz:
I have some thoughts about adding features to look up which .gitsecret directory holds the keys for which secret files. BUT -- I honestly think this PR and 'multiple .gitsecret subdirectories' will bring more confusion than clarity. 

I think separate git repos for different user permissions/groups is the way to go. But I don't want to block this feature unilaterally.

Interested to hear what @sobolevn and/or @simbo1905 think.

joshrabinowitz:
I've had a change of heart. If people want to rename the .gitsecret directory, and there's a real-world case for it, it's cool with me. We'll need to make sure it's documented and tested well.

simbo1905:
I will go along with Josh that being able to specify a different name for the configuration folder isn't a bad thing. Someone using that to workaround encrypting different sets of files for different users is perfectly fine.

On the topic of repos and environmental configuration we use a repo per environment. I think folders are not well supported as units of deployment. Many build tools don't support deployment triggers on changes to git folders. Github.com has very coarse grained access controls that are the repo level so IMHO it feels best to treat the repos as "single units of deployment". This also helps to avoid mixing up configuration and to use repo level access controls. 

Github.com does have file ownership as a concept. The idea if having a secret credential in a specific file and allowing that to be encrypted and decrypted only by the owner(s) makes a lot sense to me in terms of an enhanced access control feature for git-secret. 

Once again I vote with Josh that we shouldn't block this as someone has a use for it.

sobolevn:
I guess this feature gives some extra flexibility at a rather low cost.
So, I am all for it. Let's merge it as it is for now.

Except it should be documented and tested. 
@lrascao are you interested in providing a test for this feature after we will merge it?

sobolevn:
BTW, @joshrabinowitz @simbo1905 @hurricanehrndz @lrascao congrats with 1000 stars!
:tada:

This is huge!

joshrabinowitz:
Who wants to fix this build? I'll take a look at it tomorrow if it's still broken.

simbo1905:
hi @lrascao the CI build at 
https://travis-ci.org/sobolevn/git-secret/builds/415927649?utm_source=github_status&utm_medium=notification failed with `SC2230: which is non-standard. Use builtin 'command -v' instead.` as part of `TASK [Run lint]`/ 

Running `make lint` locally on my mac it doesn't complain about that error I suspect that the shellcheck on brew is older than the version the linux build uses. 

Do you mind making the command -v change and adding a test for your feature? 

Thanks

simbo1905:
It seems the CI build failed with a complaint about using `which` over `command -v`. 

Pulling master the files are using `command -v` https://github.com/sobolevn/git-secret/blob/b504facc5d2fa7d2c7e8c8abdee86cff5835e6f8/utils/make/make-ci.sh#L22

@lrascao this PR isn't against master its against `sobolevn:develop` which is out of date. Can you please resubmit this PR against master and add a test? 

This PR needs to be closed as its not against the master.

lrascao:
rebased onto master, we should update `CONTRIBUTING.md` though because it states:
`develop is where the development is done and the branch you should send your pull-requests to`

simbo1905:
aha. that is because we used to automatically release from master and we have recently switched it around to release when we apply a `/v.*/` tag. apologies that i hadn't see that line was there.

simbo1905:
I think the PR build is confused by the rebase as it is still awaiting a result and I see nothing on travis.

lrascao:
> Can you please resubmit this PR against master and add a test?
Regarding the tests and since this is a global option that affects all commands, what would be a good strategy?

joshrabinowitz:
I think travis doesn't see the rebase as a change, and that you have to apply one more change commit to get travis to rebuild the project

lrascao:
forced push of amended commit

sobolevn:
@lrascao thanks!

Issue #250, opened 2018-08-18
Document and test using SECRETS_DIR environment variable


joshrabinowitz:
For #247 , which supports changing the name of the .gitsecret directory via the SECRETS_DIR env var. 

@lrascao said:

> Regarding the tests and since this is a global option that affects all commands, what would be a good strategy?

joshrabinowitz:
Some ideas on how to test this feature:

Automate the test run to run every test twice, once with SECRETS_DIR set, and once without. 
*  Pros: Comprehensive. 
*  Cons: Doubles testing time.

OR

Always set the SECRETS_DIR env var in the test setup code and assume that the default .gitsecret works. 
* Pros: Simple to implement. This is probably the right thing to do first.
* Cons: We're assuming it works without setting SECRETS_DIR. 

OR

Tweak to above: Do the above test, and add a test _for each of the test files_ that specifically doesn't set the SECRETS_DIR env var. 
* Pros: Comprehensive, probably the right way to go in the long run. 
* Cons: Sort of a lot of code to write.

joshrabinowitz:
#251 adds the test "Always set the SECRETS_DIR env var in the test setup code and assume that the default .gitsecret works" described above. It also documents use of the SECRETS_DIR env var.

git-secret.io still should probably be updated to reflect SECRETS_DIR changes

joshrabinowitz:
#255 updates the gh-pages branch for <http://git-secret.io>

joshrabinowitz:
Since git-secret.io doc item moved to #250, this is complete

PR #251, opened 2018-08-25
Secrets dir docs tests 250

joshrabinowitz:
To test and document #250, which adds SECRETS_DIR env var. 
In progress.


lrascao:
👍

Issue #238, opened 2018-07-14
Git secret ignoring gpg issues with expired public key


JulianStier:
Hey,
it makes sense if git-secret would provide a warning if gpg can not be used for encryption. This happens e.g. if the public key expired. Git secret then creates an entry in the mapping configuration for the file intended for encryption, but is not able to create a file with the _.secret_ extension. This leads to an error when performing _git secret reveal_ on another repository instance.

What are the steps to reproduce this issue?
-------------------------------------------
1. Create a gpg key pair with short expiration (outdated expiration) for the public key; initialize git rep with git secret; add the identity with expired public key to git secret (tell)
2. Add a file with _git secret add_, commit, push
3. In a cloned repository, try to reveal the encrypted file

What happens?
-------------
_git secret_ gives no warning about problems with gpg

What were you expecting to happen?
----------------------------------
A warning (at least the warning given by gpg) and no entry in the mapping file.

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`) Linux - 4.4.0-130-generic #156-Ubuntu SMP Thu Jun 14 08:53:28 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

**`git-secret` path:** (`which git-secret`) /usr/bin/git-secret

**`git-secret` version:** (`git secret --version`) 0.2.4

**`git` version:** (`git --version`) 2.7.4

**Shell type and version:** (`$SHELL --version`) GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)

**`gpg` version:** (`gpg --version`) gpg (GnuPG) 1.4.20

joshrabinowitz:
Hm, I looked at this briefly and it doesn't seem easy to even detect the expiration date/time of a particular key, which seems like it'd be a pre-requisite for this feature. 

@JulianStier any tips on how we'd detect _when_ keys expire?

joshrabinowitz:
The code should use `gpg --list-keys --with-colons` for this feature.

adamgotterer:
Ran into a related problem today where someones key in the ring was expired. Spent over an hour trying to debug what happened before realizing it was an expired key. Basically hide was just exiting without any output. It would be great to know that a key was expired and that it couldn't en/decrypt.

joshrabinowitz:
@adamgotterer @JulianStier @sobolevn @simbo1905 I started working up a fix for this here:
[...]

PR #262, opened 2018-09-23
improve testing regarding expired keys and keys with only email addresses

joshrabinowitz:
For #238 and #281 . 
~~Code in progress~~
This code is ready for merge.


joshrabinowitz:
@julianStier @adamgotterer : 

Current master will issue an erorr if any file failed to decrypt (which technically fixes #238 'Git secret ignoring gpg issues with expired public key') but it won't give you any hint why.

I'm thinking of adding a patch to git-secret that informs users about keys that expire. 
My first thought is to always mention the expiration date of any keys with expiration dates. 
Any suggestions, thoughts or input on what git-secret should do about keys that have expiration times?

adamgotterer:
Awesome! Thanks for adding!

`whoknows` could also list the expirations or days until expiration by default or with a flag to show/hide. Could also have a  flag for "expiring soon".

If `hide` tells you it failed because there was an expired key(s) and lists the offender keys there could be a follow up prompt that 'Y' offers to remove those keys and continue encryption and 'N' cancels encryption.

rasendubi:
I experience a similar issue with an expired uid when I issue `git secret hide`.

git-secret calls the following command:
```sh
gpg --homedir=<repo>/.gitsecret/keys --no-permission-warning --use-agent --yes --trust-model=always --encrypt -r<uid1> -r<uid2_expired> -r<uid3_expired> -r<uid4> -o <repo>/<file>.secret <repo>/<file>
```

(I was able to fix the issue by manually `gpg --homedir=<repo>/.gitsecret/keys --edit-key <uid1>` and deleting my expired uids.)

Does this PR fixes that? I don't see any changes in `git_secret_hide.sh`. That's why I'm asking.

joshrabinowitz:
this PR is in progress.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## {{Next Version}}
+
+- Add tests for expired gpg keys, and gpg keys with only emails.
+
```
joshrabinowitz: Is it ok if we put in change information for the next changelog this way?

sobolevn: Yes, I do the same thing.

joshrabinowitz: excellent, thank you

Issue #253, opened 2018-09-05
Cannot encrypt new files unless all files are decrypted first


dave-pollock:
I was surprised to hit this issue when trying out git-secret for the first time. I'm hoping I'm just doing something dumb. Basically I cannot encrypt new files added to git-secret unless all other files in the mapping.cfg file exist. In my case they may not exist as not all contributors will be given access to private keys, but I still want them to be able to encrypt new files.

What are the steps to reproduce this issue?
-------------------------------------------
```bash
echo "Test1" > secret/file1.txt
git secret add secret/file1.txt
1 item(s) added.
git secret hide -d
done. all 1 files are hidden.
echo "Test2" > secret/file2.txt
git secret add secret/file2.txt
1 item(s) added.
git secret hide -d
git-secret: abort: file not found: secret/file1.txt
```

What happens?
-------------
I am unable to encrypt file2.txt without first decrypting file1.txt

What were you expecting to happen?
----------------------------------
I expect to be able to encrypt files even if other files which have been added to the mapping.cfg do not exist. I'd prefer to perhaps just see a warning if a file does not exist, but for git-secret to continue attempting to encrypt all other files.

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`) …
Linux 92abc9c47df7 4.9.93-linuxkit-aufs #1 SMP Wed Jun 6 16:55:56 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

**`git-secret` path:** (`which git-secret`) …
/usr/bin/git-secret

**`git-secret` version:** (`git secret --version`) …
0.2.4

**`git` version:** (`git --version`) …
git version 2.7.4

**Shell type and version:** (`$SHELL --version`) …
GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)

**`gpg` version:** (`gpg --version`) …
gpg (GnuPG) 1.4.20

joshrabinowitz:
Hm, I can see some logic why git-secret wouldn't only encrypt some of the files: The keychain might have changed since the last encryption, and if we don't re-encrypt all of the files, some files may require different keys than others to decrypt.  

@sobolevn @simbo1905 any thoughts?

joshrabinowitz:
I'm planning to have git-secret issue a more explanatory error message.

dave-pollock:
[...]

PR #263, opened 2018-09-23
Force mode for hide and reveal

joshrabinowitz:
In progress

PR #265, opened 2018-09-24
Add 'git secret hide -f' option

dave-pollock:
Adds the functionality to specify individual files when encrypting using the 'git secret hide -f' option as discussed in https://github.com/sobolevn/git-secret/issues/253. 

https://github.com/sobolevn/git-secret/issues/253

N/A

Any other comments?
-------------------
I had to mute the "run 'hide' from inside subdirectory" test when triggered from the git commit hook. This worked when I ran `make test` but not via the commit? Seems like all the tests involving subdirectories have already been muted when run via the commit hook so there may be precedence for this. If this test is supposed to work when run via the commit hook let me know..

I've also set .gitattributes to force LF line endings on .sh and .bats files for Windows devs.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 * text=auto
+*.sh text eol=lf
+*.bats text eol=lf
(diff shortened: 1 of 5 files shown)


joshrabinowitz:
We have competing PR's! See #263

Issue #176, opened 2018-05-10
`git secret tell me` does import all keys in the keychain


sobolevn:
What are the steps to reproduce this issue?
-------------------------------------------
0. Add some keys to your main `gpg` key chain
1. Create clean repo: `git init` + `git secret init`
2. Run `git secret tell me`

What happens?
-------------

```
» g tell me
done. me added as someone who know(s) the secret.
cleaning up...

~/Desktop/vue_docker  master ✗                                                 
» g whoknows
mail@sobolevn.me
no-reply@sobolevn.me
some@mail
...
ja@...
ev@...
ev@...
me@...
```

What were you expecting to happen?
----------------------------------

There should be an error, that this key does not exist.

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`) latest `macOS` 

**`git-secret` path:** (`which git-secret`) `/usr/bin/local/git-secret`

**`git-secret` version:** (`git secret --version`) `master` (commit hash: `commit dff1aaf9242b25826d3db77868ba5508d5ac48e5`)

**`git` version:** (`git --version`) git version 2.16.1, hub version 2.2.9

**Shell type and version:** (`$SHELL --version`) `zsh 5.3.1 (x86_64-apple-darwin15.6.0)`

**`gpg` version:** (`gpg --version`) `gpg (GnuPG) 2.2.3`

joshrabinowitz:
Shoot, this seems like an issue related to gpg 2.2.3 

In 2.2.6 with dff1a I get: 

````
% git secret tell NOSUCHFILE
gpg: WARNING: nothing exported
git-secret: abort: no keyfile found for 'NOSUCHFILE'. Check your key name: 'gpg --list-keys'.
cleaning up...
````

**Comment by joshrabinowitz** [comment]
Assuming you're using gpg from brew, can you try upgrading to gpg 2.2.6?

**Comment by joshrabinowitz** [comment]
So, the bug here is that `git secret tell substr` will add all keys with a uid that contain substr _anywhere in the uid_.

**Comment by joshrabinowitz** [comment]
@sobolevn can you confirm that #243 this fixes this issue?

**Comment by sobolevn** [comment]
@joshrabinowitz still happens to me with the `master` branch (`commit 6251fae396da5dc181f5a015b2cb5654616475dc (HEAD -> master, origin/master, origin/HEAD)`):

```
~/Desktop/git-secret-test  master ✗                                            
» g init
[...]

PR #267, opened 2018-09-25
make sure email addresses used with 'tell' and 'killperson' exist in keyring

joshrabinowitz:
This PR a fix for #176. 

#176 is a bug with `tell` which adds all keys with a substring matching the specified username. For example, `git secret tell bob` will add all keys with the string `bob` anywhere in the username or email address. 

Here's the details:
 * each gpg key is associated with an `email address`, and optionally a `name`. 
 * gpg does substring matching against both the `email address` and the `username` when looking for keys matching a string.

Also, 
 * the `name`associated with a key is not guaranteed to be unique and 
 * the `email` can be a substring of other names or emails

For the above reasons, this PR changes the code so you have to specify users by email address with `git secret tell`. (Edit: The docs always said to use email addresses so there is no doc change.)

@sobolevn @simbo1905 thoughts/suggestions/input?


joshrabinowitz:
This PR does still have a bug where if 
* one email address is a substring of another (for example bob@github.com and joebob@example.com), and 
* you use `git secret tell bob@github.com`, 
then it will 
* add keys for  both users and not just the one you probably meant. 

I don't see an easy way to fix that and figured we could tackle that in a separate PR. If you see the right way to fix this issue let me know @sobolevn @simbo1905  @dave-pollock

src/commands/git_secret_tell.sh
```diff
@@ -64,6 +64,21 @@ function tell {
     _abort "you must provide at least one email address."
   fi
 
+  # TODO: this block is duplicated in git_secret_killperson.sh and should be factored
```
sobolevn: Should we refactor it now or you feel like it's not the time yet?

src/commands/git_secret_killperson.sh
```diff
@@ -24,10 +24,25 @@ function killperson {
...
+  local gpg_uids
+  gpg_uids=$(_get_users_in_gpg_keyring "$secrets_dir_keys")
+  for email in "${emails[@]}"; do
+    local email_ok=0
+    for uid in $gpg_uids; do
+        if [[ "$uid" == "$email" ]]; then
+            email_ok=1
+        fi
+    done
+    if [[ "$email_ok" == 0 ]]; then
```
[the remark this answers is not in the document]
joshrabinowitz: will revisit, thanks @simbo1905

src/commands/git_secret_tell.sh
joshrabinowitz: I'll take another look and let you know @sobolevn

src/commands/git_secret_killperson.sh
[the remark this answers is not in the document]
joshrabinowitz: fixed!

src/commands/git_secret_tell.sh
joshrabinowitz: factored.

Issue #266, opened 2018-09-24
Add per repository config option for SECRET_EXTENSION


gitmalet:
It would be great to have an option in some config file (preferably a tracked file inside `.gitsecrets`) to set the `SECRET_EXTENSION` for this repository.

What were you expecting to happen?
----------------------------------
The extension for git-secret managed files is set to the configured extension for the specific repository, overriding the environment variable.

sobolevn:
@archmalet that's an interesting idea indeed! 

For example, for the next project I have decided to use custom secret files' extension.
So, now all other developers should also specify this variable somewhere in the code/docs/build tools. Keeping inside our own `.gitsecrets/` directory might help.

robinleander:
Interesting proposal! I've went ahead and implemented a possible solution for this!

Please tell me what you think!

PR #270, opened 2018-09-28
Implement per repository SECRETS_EXTENSION setting

robinleander:
Allows setting `$SECRETS_EXTENSION` via `.gitsecret/settings.cfg` in this manner:

```
SECRETS_EXTENSION=".gpg"
```

Yes, #266.

```
66 tests, 0 failures
```

Any other comments?
-------------------

Please tell me what you think!

The implementation sources the given settings file. It uses a fairly tight regular expression to get the correct setting and validate that no malicious code is executed.
Are there better ways to implement this?

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -12,8 +12,14 @@ _SECRETS_DIR_KEYS_MAPPING="${_SECRETS_DIR_KEYS}/mapping.cfg"
 _SECRETS_DIR_KEYS_TRUSTDB="${_SECRETS_DIR_KEYS}/trustdb.gpg"
 
 _SECRETS_DIR_PATHS_MAPPING="${_SECRETS_DIR_PATHS}/mapping.cfg"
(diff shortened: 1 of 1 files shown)


joshrabinowitz:
Howdy @ambiso! Thanks for this PR, but it can't be merged because it's failing the Travis CI builds.

The issue seems to be a lint/shellcheck failure, looks like it's: https://github.com/koalaman/shellcheck/wiki/SC1090 . 
Let us know if you have any questions.

joshrabinowitz:
Users can already change the SECRETS_EXTENSION env variable to change the extension git-secret uses on the hidden (encrypted) files. 

Do we really want to support a .gitsecret/settings.cfg file? If so, it should probably be supported for it to change any SECRETS_ configuration. 

I think this feature is overkill, and we should not support .gitsecret/settings.cfg without more thought/discussion. Also this PR has languished for 2 months without changes. 

I suggest we close this PR without merging.

@sobolevn @ambiso @simbo1905 thoughts?

Issue #283, opened 2018-11-30
Feature request: better messaging about expired keys


joshrabinowitz:
If gpg encounters an expired key while being used in git-secret, it can issue confusing error messages.

It would be nice to directly detect expired keys.

People have also requested features to show and maybe even warn users about keys and their expiration dates.

(_Edit: all above issues also apply to revoked keys_ EditToTheEdit: Except key revocation requires network access and so should be handled differently and in a separate issue as we may not even have network access when running git-secret)

See also: #281 regarding expired keys.

From PR #262:

joshrabinowitz wrote:
>Current master will issue an erorr if any file failed to decrypt (which technically fixes #238 'Git 
>secret ignoring gpg issues with expired public key') but it won't give you any hint why.

>I'm thinking of adding a patch to git-secret that informs users about keys that expire.
>My first thought is to always mention the expiration date of any keys with expiration dates

adamgotterer wrote:
>whoknows could also list the expirations or days until expiration by default or with a flag to 
>show/hide. Could also have a flag for "expiring soon".

joshrabinowitz:
* whoknows now has a -l option which shows key expirations.
* we should probably still detect when we attempt to use expired (or 'soon' expiring) keys so we can tell user

joshrabinowitz:
See also: #290, 'cannot encrypt for a certain set of keys'

ajayramezh:
Any latest update on this feature, I am planning to work on this.

joshrabinowitz:
Hello, @ajayramezh:
There are no more recent updates, the last updates to this topic were in PRs #626 and #287. 
We look forward to seeing your changes.

joshrabinowitz:
Closed as 'wontfix', see #516

joshrabinowitz:
Since it looks like we're going to merge  #553, I'm re-opening this because users are likely to be confused if `git secret tell` silently ignores expired keys.

joshrabinowitz:
Note that merged PR #518 for issue #516 shows error messages from gpg when it returns an error.

#553 also shows better messaging about expired keys by warning when unusable keys are skipped.

joshrabinowitz:
Closed by #553

PR #287, opened 2018-12-02
WIP: test with expired key, add 'whoknows -l'

joshrabinowitz:
Regarding #238 (Git secret ignoring gpg issues with expired public key) 
and #283 (Feature request: better messaging about expired keys)

This PR has code to test for and show data about expired keys:

* adds -l option,  'whoknows -l' to show key expiration dates
* adds docs and tests for `whoknows -l`,
* adds tests for expired keys,
* adds epoch_to_date functions

~~WIP,~~ Input welcome

### DIFF
--- a/src/_utils/_git_secret_tools_osx.sh
+++ b/src/_utils/_git_secret_tools_osx.sh
@@ -36,7 +36,7 @@ function __epoch_to_date_osx {
     #date -r 234234234 +"%Y-%m-%d"
     local cmd="/bin/date -r $epoch +'%Y-%m-%d'"
     local datetime
-    datetime=$($cmd)
+    datetime=$("$cmd")
     echo "$datetime"
   fi
 }
--- a/src/_utils/_git_secret_tools_osx.sh
+++ b/src/_utils/_git_secret_tools_osx.sh
@@ -27,3 +27,16 @@ function __get_octal_perms_osx {
   # see _git_secret_tools_freebsd.sh for more about stat's format string
   echo "$perms"
 }
+
+function __epoch_to_date_osx {
+  local epoch=$1;
+  if [ -z "$epoch" ]; then
+    echo ''
+  else
+    local cmd="/bin/date +%F -r $epoch"
+    local datetime
+    datetime=$($cmd)
+    echo "$datetime"
+  fi
+}
+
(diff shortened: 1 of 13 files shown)


joshrabinowitz:
@adamgotterer @julianstier @sobolevn @simbo1905 input welcome

src/_utils/_git_secret_tools_osx.sh
```diff
@@ -27,3 +27,16 @@ function __get_octal_perms_osx {
...
 }
+
+function __epoch_to_date_osx {
+  local epoch=$1;
+  if [ -z "$epoch" ]; then
+    echo ''
+  else
+    #date -r 234234234 +"%Y-%m-%d"
+    local datetime
+    datetime=$(/bin/date -r "$epoch" +'%Y-%m-%d')
```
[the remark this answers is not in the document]
joshrabinowitz: I had been following the example of `shasum`, fullpath is not needed.

src/main.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-# we don't use set -e, it messes up hide -F and reveal -F error handling
+set -e
```
joshrabinowitz: I had previously removed `set -e` from `src/main.sh`, because otherwise issues with (for example) expired keys would cause the script to stop here.

I've re-enabled `set -e` here, and disabled the `set -e` feature selectively, just around the `gpg` use I want to capture issues with. (You can see the changes in `commands/git_secret_hide.sh` and `src/_utils/_git_secret_tools.sh`.)

Is there a better way to do this?

src/commands/git_secret_hide.sh
```diff
@@ -160,11 +160,17 @@ function hide {
   
       # encrypt file only if required
       if [[ "$fsdb_file_hash" != "$file_hash" ]]; then
+
+        set +e   # disable 'set -e' so we can capture exit_code
```
joshrabinowitz: We disable 'set -e' here and re-enable it below.

src/main.sh
joshrabinowitz: interested to hear your thoughts @sobolevn

joshrabinowitz:
I'm ready to (squash and) merge this if @sobolevn is happy with it

src/commands/git_secret_hide.sh
sobolevn: Yes, looks fine!

sobolevn:
Squashed and merged! Thanks @joshrabinowitz!

Issue #289, opened 2018-12-04
Reach out to package maintainers


sobolevn:
There are some packages that are maintained by other people.
We should reach to them and invite to collaborate.

Originally posted as: https://github.com/sobolevn/git-secret/issues/284

joshrabinowitz:
Maybe the easiest way to reach them is to put notes to them in the README.md or similar and releasing a new version of git-secret.

sobolevn:
@joshrabinowitz that's a nice idea!

joshrabinowitz:
See also #312, #315, #316 .

PR #301, opened 2019-01-04
added notes about packages and for package maintainers

joshrabinowitz:
As per #289 "Reach out to package maintainers"


CONTRIBUTING.md
```diff
@@ -98,7 +98,7 @@ Here are some links to gnupg documentation that might be useful for those workin
 
 #### Travis releases
 
-When creating a commit inside `master` branch, `travis` on successful build will publish new `deb` and `rpm` packages to [`bintray`][bintray].
+It was once the case that creating a commit inside `master` branch, `travis` on successful build should publish new `deb` and `rpm` packages to [`bintray`][bintray].
```
sobolevn: We should also note that now it creates a release on tag. At least it should do so.

joshrabinowitz: ok I'll add that soon

joshrabinowitz: It now says:

    After you commit a tag that matches the pattern '^v.*$' and travis succeeds, 
    Travis should publish new `deb` and `rpm` packages to [`bintray`][bintray].

PR #313, opened 2019-01-16
remove unused functions

joshrabinowitz:
Removes unused functions from  src/_utils/_git_secret_tools.sh


joshrabinowitz:
@sobolevn I'll plan to merge this next week unless there's resistance.

PR #321, opened 2019-01-19
Add verbose option to  'git secret tell' showing output of key imports

joshrabinowitz:
as per Alpine Linux 'testing' patch here: https://github.com/alpinelinux/aports/blob/fb9142d5f0df59f9a0e381e6b79829dd6b40c680/testing/git-secret/dont-hide-gpg-output.patch

Their patch to 'src/_utils/_git_secret_tools.sh' is already in master.

### DIFF
--- a/src/commands/git_secret_tell.sh
+++ b/src/commands/git_secret_tell.sh
@@ -95,7 +95,9 @@ function tell {
     # Importing public key to the local keychain:
     local secrets_dir_keys
     secrets_dir_keys=$(_get_secrets_dir_keys)
-    $SECRETS_GPG_COMMAND --homedir "$secrets_dir_keys" --no-permission-warning --import "$keyfile" > /dev/null 2>&1
+    # we used to discard output from the below command but include it now for debug-ability
+    # as per patch from Alpine Linux
+    $SECRETS_GPG_COMMAND --homedir "$secrets_dir_keys" --no-permission-warning --import "$keyfile"
     exit_code=$?
     if [[ "$exit_code" -ne 0 ]]; then
       _abort "problem importing public key for '$email' with gpg: exit code $exit_code"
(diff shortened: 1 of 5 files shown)


joshrabinowitz:
@simbo1905 @sobolevn any opinion on this PR? 

It shows gpg's output from key imports (instead of redirecting them to /dev/null) in order to increase debuggability as per alipinelinux's patch to git-secret 0.2.2

sobolevn:
Their patch also covers other different commands.
I would say that this is normally handled via `verbose` level.

And it should be configurable. @joshrabinowitz do you agree?

joshrabinowitz:
When I looked at their patches, I found that the other diffs in that patch (to src/_utils/_git_secret_tools.sh) have been already independently applied to git-secret.

joshrabinowitz:
As for the -v option for verbosity in `git secret tell`: it feels like creeping featuritis a little bit but otherwise it's totally consistent with everything about git-secret. So I'll give it a +1.

src/commands/git_secret_tell.sh
```diff
@@ -21,13 +21,16 @@ function tell {
...
+  local verbose=0
 
   # A POSIX variable
   # Reset in case getopts has been used previously in the shell.
   OPTIND=1
 
-  while getopts "hmd:" opt; do
+  while getopts "vhmd:" opt; do
     case "$opt" in
+      v) verbose=1;;
```
sobolevn: Maybe we can make `-v` a global option? Something like:

```bash
git secret -v tell user@email.com
```

or 

```bash
GIT_SECRET_VERBOSE=1 git secret tell some@random.name 
```

joshrabinowitz: We could.

Also presumably since it would also effect `clean` and `hide`'s -v options too, I suggest we implement GIT_SECRET_VERBOSE (if we decide to do it) in a separate PR. Discussion moved to #323

Issue #335, opened 2019-01-30
Fix issue installing fpm with kitchen on alpine and redhat-based setups


joshrabinowitz:
Multiple configurations fail with the same error (newlines and bold emphasis added):

fatal: [localhost]: FAILED! => {"changed": false, "cmd": "/usr/bin/gem install --no-user-install --no-document fpm", "msg": "ERROR:  Error installing fpm:\n\tERROR: Failed to build gem native extension.\n\n    current directory: /usr/local/share/gems/gems/childprocess-1.0.0/ext\n/usr/bin/ruby mkrf_conf.rb\n\ncurrent directory: /usr/local/share/gems/gems/childprocess-1.0.0/ext\nrake RUBYARCHDIR=/usr/local/lib64/gems/ruby/childprocess-1.0.0 RUBYLIBDIR=/usr/local/lib64/gems/ruby/childprocess-1.0.0\nsh: rake: command not found\n\nrake failed, exit code 127\n\nGem files will remain installed in /usr/local/share/gems/gems/childprocess-1.0.0 for inspection.\nResults logged to /usr/local/lib64/gems/ruby/childprocess-1.0.0/gem_make.out",

 "rc": 1, 

"stderr": "ERROR:  Error installing fpm:\n\tERROR: Failed to build gem native extension.\n\n    current directory: /usr/local/share/gems/gems/childprocess-1.0.0/ext\n/usr/bin/ruby mkrf_conf.rb\n\ncurrent directory: /usr/local/share/gems/gems/childprocess-1.0.0/ext\nrake RUBYARCHDIR=/usr/local/lib64/gems/ruby/childprocess-1.0.0 RUBYLIBDIR=/usr/local/lib64/gems/ruby/childprocess-1.0.0\nsh: rake: command not found\n\nrake failed, exit code 127\n\nGem files will remain installed in /usr/local/share/gems/gems/childprocess-1.0.0 for inspection.\nResults logged to /usr/local/lib64/gems/ruby/childprocess-1.0.0/gem_make.out\n", 

[...]

PR #336, opened 2019-01-30
Fix kitchen builds, document dev setup on CentOS

joshrabinowitz:
* records changes I had to make to get kitchen working on CentOS 7
* had to make sure rake was installed, or encountered failure installing 'fpm' (See #335)

### DIFF
--- a/.ci/integration/tasks/dependencies.yml
+++ b/.ci/integration/tasks/dependencies.yml
@@ -39,9 +39,8 @@
 #    user_install: no
 
 # so we install rake and fpm together at once with gem.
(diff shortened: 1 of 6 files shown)


joshrabinowitz:
This also fixes builds on fedora and alpine linux. 

Debian based builds are currently failing during install of fpm.

joshrabinowitz:
fixes for Debian/ubuntu applied too

joshrabinowitz:
The build changes in this PR are unneeded now. 
The underlying problem appears to have been a transient upstream issue. 
Doc changes to be moved to separate PR.

PR #338, opened 2019-01-30
Upgrade ansible

joshrabinowitz:
upgrades ansible in builds from 2.5.0 to current 2.6.12


joshrabinowitz:
This does not actually appear to upgrade ansible in builds. Closing.

PR #358, opened 2019-02-14
Added -g flag to git secret changes #340

krufab:
Adds -g flag to git secret changes to output conflicts in git style.
Described in PR #340

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -711,3 +711,62 @@ function _decrypt {
   # at this point the file should be written to disk or output to stdout
 }
 
+# usage: _check_if_plaintexts_have_conflicts FULL_PATH_TO_FILE1 FULL_PATH_TO_FILE2 ...
+#    ie: _check_if_plaintexts_have_conflicts $file1 $file2 /full/path/to/the/file3
+#    ie: _check_if_plaintexts_have_conflicts "${to_show[@]}"
+# (checks if a list of file have conflicts. If there are one or more conflict,
+# it terminates the execution with an error message)
+function _check_if_plaintexts_have_conflicts {
+  # required:
+  local to_show=("${@}")
+
+  local conflicting=()
+
+  for line in "${to_show[@]}"; do
+    local filename
+    local path
+
+    filename=$(_get_record_filename "$line")
+    path=$(_append_root_path "$filename")
+
+    if [[ -f "$path" ]]; then
+      local has_conflict
+      has_conflict=$(_check_if_plaintext_has_conflicts "${path}")
+      if [[ "1" == "${has_conflict}" ]]; then
+        conflicting+=("${path}");
+      fi
+    fi
+  done
+
+  if [[ ${#conflicting[@]} -gt 0 ]]; then
+      echo -e "Conflicts were found in the following file(s):\n"
+      printf '%s\n' "${conflicting[@]}"
+      echo -e "\nResolve conflicts before continuing. Aborting."
+      exit 1
+  fi
+}
+
+# usage: _check_if_plaintext_has_conflicts FULL_PATH_TO_FILE
+#    ie: _check_if_plaintext_has_conflicts /full/path/to/the/file
+#    ie: _check_if_plaintext_has_conflicts "$file"
+# (checks if a single file contains conflicts)
+# Returns 1 on conflicts, otherwise 0
+function _check_if_plaintext_has_conflicts {
+  local path="${1}"
+  set +e
+  if grep -E "^<<<<<<< file-on-disk$|^>>>>>>> file-from-secret$" "${path}" > /dev/null; then
+    echo 1
+  else
+    echo 0;
+  fi
+  set -e
+}
+
+# usage: check_if_busybox_diff
+# diff -D is not supported by diff applet
+function _check_if_busybox_diff {
+  if diff 2>&1 | grep BusyBox > /dev/null; then
+    local msg="diff -D is not supported by BusyBox diff applet. Please install diffutils (very likely via apk add diffutils)"
+    _abort "$msg"
+  fi
+}
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Folders:
-_SECRETS_DIR=${SECRETS_DIR:-".gitsecret"}   
+_SECRETS_DIR=${SECRETS_DIR:-".gitsecret"}
 # if SECRETS_DIR env var is set, use that instead of .gitsecret
 # for full path to secrets dir, use _get_secrets_dir() from _git_secret_tools.sh
 _SECRETS_DIR_KEYS="${_SECRETS_DIR}/keys"
@@ -87,6 +87,10 @@ AWK_GPG_VER_CHECK='
 }
 '
 
+MARKER_FILE_ON_DISK="<<<<<<< file-on-disk"
+MARKER_SEPARATOR="======="
+MARKER_CONTENT_FROM_SECRET=">>>>>>> content-from-secret"
+
 # This is 1 for gpg version 2.1 or greater, otherwise 0
 GPG_VER_21="$(gpg --version | gawk "$AWK_GPG_VER_CHECK")"
 
@@ -116,7 +120,7 @@ function _os_based {
     Linux)
       "$1_linux" "${@:2}"
     ;;
-	
+
     MINGW*)
       "$1_linux" "${@:2}"
     ;;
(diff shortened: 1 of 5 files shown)


src/_utils/_git_secret_tools.sh
```diff
@@ -711,3 +711,62 @@ function _decrypt {
...
+
+# usage: _check_if_plaintext_has_conflicts FULL_PATH_TO_FILE
+#    ie: _check_if_plaintext_has_conflicts /full/path/to/the/file
+#    ie: _check_if_plaintext_has_conflicts "$file"
+# (checks if a single file contains conflicts)
+# Returns 1 on conflicts, otherwise 0
+function _check_if_plaintext_has_conflicts {
+  local path="${1}"
+  set +e
+  if grep -E "^<<<<<<< file-on-disk$|^>>>>>>> file-from-secret$" "${path}" > /dev/null; then
```
joshrabinowitz: What is going on here? Can we simplify/clarify this?

src/commands/git_secret_changes.sh
```diff
@@ -60,10 +65,28 @@ function changes {
     decrypted="${decrypted_x%x*}"
     # we ignore the exit code because _decrypt will _abort if appropriate.
 
+    if [[ 1 -eq $git ]]; then
+      _check_if_busybox_diff
+      # We output a merged text where the diffs are surrounded by git-style conflicts
+      diff_x=$(
+        diff -D AAAAAAA "${path}" <(echo -n "$decrypted") | \
```
joshrabinowitz: Ditto my comment above about simplicity/clarity

tests/test_changes.bats
```diff
@@ -7,7 +7,8 @@ SECOND_FILE_TO_HIDE="$TEST_SECOND_FILENAME"
 THIRD_FILE_TO_HIDE="$TEST_THIRD_FILENAME"
 FILE_NON_EXISTANT="NO-SUCH-FILE"
 FILE_CONTENTS="hidden content юникод"
-
+FILE_CONTENTS_UPDATED="hidden content юникод\na_new_line"
+FILE_CONTENTS_CONFLICTS="hidden content юникод\n<<<<<<< file-on-disk\na_new_line\n=======\n>>>>>>> file-from-secret"
```
joshrabinowitz: ditto here. Not sure we need file-on-disk/file-from-secret?

joshrabinowitz: Also I don't think grep -e is guaranteed to exist (though it seems to on all our tests systems)

joshrabinowitz:
I looked at what change this makes to `git changes`, here's what I found:

````
%  git changes 'space file'
changes in /tmp/space file:
--- /dev/fd/63	2019-02-15 16:38:14.749001940 -0500
+++ "/tmp/space file"	2019-02-15 16:38:14.621002146 -0500
@@ -1 +1,2 @@
 hidden content юникод
+new content

% git changes -g 'space file'
hidden content юникод 
<<<<<<< file-on-disk
a_new_line
=======
>>>>>>> file-from-secret
````

The goal of this PR seems to be to change how 'git changes' shows the diff. 
I can see how it'd be nice to replace `/dev/fd/63` with something more expressive, and the date strings are not needed.

But can we do something more standard and with less code?

src/commands/git_secret_changes.sh
krufab: This is the part which generates the conflicts blocks (git style) in case files are different (it outputs the full file if there are no changes).
According to the man page, `diff -D NAME  --ifdef=NAME  Output merged file to show `#ifdef NAME' diffs.`, that is diff will mark the begin and end of each change with C style ifdef / else / ifndef pre-compiler style blocks. AAAAAAA is the text/variable you want to use as marker. This is pretty nice, but as we are using git secret with, well, git, it would be nicer to have a git style markers.
For this reason, I pipe the output of the diff -D to multiple sed commands to remove the C style ifdef etc and substitute them with git style >>>>>>> ====== <<<<<< markers.
Normally, git would compare versions of a file, so you would have a commit hash or branch name, but we don't have this info, and the file name is the same, so I thought it was clearer to have the conflicts clearly state what's on disk and what's from the secret.

tests/test_changes.bats
krufab: The FILE_CONTENTS_CONFLICTS is a fixture containing the expected output of changes -g when there are conflicts

krufab:
Well, partially. The goal is to extend (it's a new flag, not the default behaviour) the way git secret changes shows the diff. The default is a generic diff change, which is ok, but I personally find it hard to read, with a bit too much noise (+-@ signs and specific padding) and not well integrated with git (generally, git refuses to commit files with non resolved conflicts, but it would have no problem to commit files with default diff markers).
Therefore I came up with this idea of having all differences visually marked with git-style conflicts.

Git has a tool to generate the conflicts files, but unfortunately it also automatically merges non conflicting changes (based on the history). This is a behaviour I don't want to use for secret files, which unfortunately are not tracked (at least the plaintexts). Therefore the default git tool might apply changes, which we would fail to recognise, because the plaintext doesn't show up in git diff / git status. I prefer to have the user manually approve conflicts and manually pick the changes from the secret or the actual file on disk.

I would have liked to have a more elegant one-liner to substitute the markers, but I wasn't able to use sed with variables and \n, so I had to use several pipes.

tests/test_changes.bats
joshrabinowitz: yes, of course. I was commenting more on the content of the string, and if we really want/need those strings in the output,

joshrabinowitz:
@krufab, apologies for the bluntness, but I'm not inclined to merge this PR: I like the idea and feature request, but I don't love the implementation. I'd like to see something much simpler/easier to explain, and with much less code. Can we re-think this feature to be simpler, easier to explain, and a much smaller PR?

src/commands/git_secret_changes.sh
joshrabinowitz: I would really prefer if we didn't fetch one diff format and then convert it to another.

krufab:
Well, the git style diff is exactly one liner, I don't think we can get much shorter than that.

Sed substitution with variables and "\n" does not seem to be very portable: all the examples I found on stack overflow failed on my mac. Or at least I failed to code it correctly. Therefore I had to hardcode all the placeholders and markers. Not the best approach, but works on all the platforms tested.
The conflict check functions are a bit prolix, but in line with the rest of the code, I think.

git secret hide will only use the same check functions and git secret reveal will use the git secret changes -g to output the file, so there won't be much more changes.

joshrabinowitz:
I was just looking at https://pubs.opengroup.org/onlinepubs/009695399/utilities/diff.html and read about the -e option:

`
-e : Produce output in a form suitable as input for the ed utility, 
which can then be used to convert file1 into file2.
`

Perhaps we could use 'diff -e' (with all the appropriate flags etc) and then pipe it into `ed` to create the 'merged-with-conflicts' type output you're seeking? Or perhaps we create a unified diff and then pipe it into 'patch' to create the merge? I'll admit I'm not sure what would happen with conflicts in these case, but we'd be using unix tools to do more of the work so that would be nice.

krufab:
I checked diff -e but i don't think it's gonna be any easier or simpler. The output is ed-specific sequence of commands to go from file a to file b. To add git-style markers, you'd need anyway to parse diff -e output and manipulate the 'c', '.', 'd' commands with the conflict markers.

I checked the other flags available in diff, but they don't seem to be any any better than -D.

In reality, I'd have like to use some git command (i.e. [git-merge-file](https://git-scm.com/docs/git-merge-file)), but it also merges the changes (and would need the history of a file, which does not exist for plaintexts), while I need to manually decide on each difference.

I have updated the code, making the sed sequence a bit more readable, moving code to constants and adding comments. Better than that, I'm afraid I can't do.

joshrabinowitz:
How about scaling back to first just a flag that would detect conflicts? 
Is there a simple way to detect that condition in code?

krufab:
@joshrabinowitz,
I'm afraid I cannot help with that.
I like this project and since we introduced it in our cicd process, it made few things easier (we can version configurations) but on the other hand the process is a real nightmare. It's too difficult to keep track of configurations changing in multiple branches and it's too prone to errors. I can use it, but my colleagues are not really inclined to deal with it, so for me it became a burden.

This is why I implemented the safe reveal and hide and the git-style syntax for conflict. I've created the original PR, with working code. I updated it according to the PR comments, adapting it to the style and logic of this project, breaking it into steps and also cleaning up the code as best as I could. Now I need to use it and I'm not really willing to make more changes to improve the code quality.
This PR seems to be working, has non failing test and I believe it can be merged and released. If the code really needs to be polished, or needs some better logic, you can open an issue and see if anyone is able to solve it.
If nothing changes, I think I will merge the code in my own repository over the weekend, as next week I'll need to use it.

For reference, have a look at this [question](https://stackoverflow.com/questions/54346394/merge-2-text-files-with-differences-marked-as-conflicts) on stack overflow, which I've opened before implementing the code. Diff -D + sed is the most suggested option.

BTW, to detect conflicts, it's enough to use git secret changes as it is now.

joshrabinowitz:
Thanks for working on this and for the discussion. We appreciate your PR, your code, and the idea. 

Would it make sense to put your features into an external git-secret plugin like 
https://github.com/sobolevn/git-secret/wiki/Third-party-plugins ?

I say this because it would be great not to have two different versions of git-secret floating around being different.

joshrabinowitz:
Not merging this PR as-is

PR #369, opened 2019-03-01
Add sed regex to _get_git_root_path that changes <driveletter>:/ to /…

Mohjive:
Under git for windows `git rev-parse --show-toplevel` outputs paths in the form **C:/path/to/repo**. This breaks gpg with the --homedir set to this path as no leading slash is considered a relative path and stacks the argument with current directory, leading gpg to trying to access **/c/path/to/repo/C:/path/to/repo/.gitsecret**.
The suggested change fixes this by patching the `_get_git_root_path` to change the driveletter part to an absolute *nix path, which will then work with `gpg --homedir`

No open issues. Did PR instead of submitting one.

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -141,7 +141,7 @@ function _os_based {
 
 function _clean_path {
   # This function transforms windows paths to *nix paths
-  echo $1 | sed 's#^\([a-zA-Z]\):/#/\1/#'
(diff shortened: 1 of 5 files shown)


joshrabinowitz:
@Mohjive I'm working up a PR that lets us test git-secret on Travis-CI windows environments so we can show this failure in the current code base and that your PR fixes the issue. 

Is there any possibility you can take a look at #370 and see if you can tell what's wrong on windows? Would be much appreciated.

joshrabinowitz:
Closing -- this code is all in #375

Issue #40, opened 2016-06-25
windows support


sobolevn:
Right now `git-secret` does not support `windows` operating systems. But the architecture is suitable to handle it.

I have never actually worked with `windows`. And I have neither time nor will to get starting.

What are the limitations? 
What frameworks/external programs does it require to run `bash` on `windows`? 
Is it even possible?
How many tests do pass right now (if any)?

I would highly appreciate any help on this topic!

timchurchard:
On Windows 10 using Windows Subsystem for Linux (Beta) (Based on Ubuntu 14.04) git-secret worked out of the box 100% test pass after installing git and build-essential.
https://msdn.microsoft.com/en-gb/commandline/wsl/install_guide

I'll try in Windows 7 VM using mingw tomorrow.  http://www.mingw.org/

sobolevn:
Wow! Thats great! Are there any ways to automate this testing? 
Some kind of  Travis for Windows?

timchurchard:
I've tried using Cygwin https://cygwin.com on Windows 7 but the tests are failing, I think cygwin is having trouble with my username "First Last".  I'll continue trying.

For automated test how about http://docs.tea-ci.org -- Although I haven't used it.  It's free for open source projects.

sobolevn:
Hm, maybe https://www.appveyor.com/ is better? 
I have never actually used neither of these.

As I see it, TeaCI is a fork of another project (drone.io). The master branch was not updated for 9 months for now.

beInDev:
I had no problem running the tests on cygwin.

Had bats installed and recognized in path. 

`git secret` also seemed to work as expected (haven't tried extensively yet), only had to change [_git_secret_tools.sh](/sobolevn/git-secret/blob/master/src/_utils/_git_secret_tools.sh#46) and uncomment the windows os check case and made it so it would use the [_git_secret_tools_linux.sh](/sobolevn/git-secret/blob/master/src/_utils/_git_secret_tools_linux.sh)

**OS:** Windows 10
**Cygwin version:** CYGWIN_NT-10.0 2.8.0(0.309/5/3) x86_64 Cygwin

### Test Results:
![image](https://cloud.githubusercontent.com/assets/2298687/25714100/3b1aeac2-30f7-11e7-8d85-663eed9878f5.png)

sobolevn:
@beInDev wow that's very cool! 

What's you `gpg`, `bash` and `git` version?
Are you interested in making a PR with:

1. Your changes in code
[...]

PR #370, opened 2019-03-01
simple test for travis-ci windows support

joshrabinowitz:
experiment with travis-ci and windows

### DIFF
--- a/.ci/before_script.sh
+++ b/.ci/before_script.sh
@@ -30,13 +30,12 @@ fi
 # Windows
 # msys2 needed for rm?
 if [[ "$GITSECRET_DIST" == "windows" ]]; then
-  choco install msys2 gawk gnupg-modern make -y
-  #choco install make -y
+  choco install wsl gawk gnupg-modern make -y
 
-  echo -n "location of bash: "
+  echo -n "before_script.sh: location of bash: "
   which bash || true
 
-  echo -n "location of env: "
+  echo -n "before_script.sh: location of env: "
   which env || true
 fi
 
--- a/.ci/before_script.sh
+++ b/.ci/before_script.sh
@@ -29,7 +29,8 @@ fi
 
 # Mac:
 if [[ "$GITSECRET_DIST" == "windows" ]]; then
-  choco install gawk gnupg-modern make
+  choco install gawk gnupg-modern make msys2 
+  # msys2 provides bash and "is a software distro and building platform for Windows."
 fi
 
 # Linux:
--- a/.ci/before_script.sh
+++ b/.ci/before_script.sh
@@ -32,10 +32,12 @@ if [[ "$GITSECRET_DIST" == "windows" ]]; then
   choco install gawk gnupg-modern make msys2 
   # msys2 provides bash and "is a software distro and building platform for Windows."
 
-  refreshenv
+  # refreshenv
   # run refreshenv because choco says:
   #  "Environment Vars (like PATH) have changed. Close/reopen your shell to
   #  see the changes (or in powershell/cmd.exe just type `refreshenv`).
+  # BUT when you run it, travis errors with:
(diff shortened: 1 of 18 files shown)


joshrabinowitz:
@sobolevn @Mohjive @simbo1905 

I've been trying to set up this test of git-secret on travis-ci's windows build environment <https://docs.travis-ci.com/user/reference/windows/> but keep having issues executing scripts with `bash` (see PR). Help?

I'd like to get git-secret windows builds going for windows PRs , tests and possibly releases. Please take a look a the PR and the travis-ci windows output and see if you can spot anything wrong. Thank you.

Mohjive:
@joshrabinowitz I did try to run the tests the other day and one issue I found is that `ps` has incompatible options under git for windows version of bash.
The other was that I had to install **bats** (installed it from git).

Mohjive:
and what I can see from the travis docs it uses git bash, so it should use the same environment (and no need to install bash separately).

sobolevn:
Sorry, @joshrabinowitz, but I do not have any experience with Windows (I am lucky man).
Maybe we can ask for help in https://travis-ci.community/c/environments/windows ?

Mohjive:
It’s impossible to follow any process. Make it work on your machine with git for windows first and make that one commit and **then** push that one change. I cannot give any feedback on these commits, because the signal to noise ratio is way too low.

joshrabinowitz:
@Mohjive Thanks for taking a look at the PR.

To raise the signal: the current problem has to do with subshells finding various commands at run time; all the recent changes to this branch have been around the PATH env variable. You can see the problems, which all relate to this, at the end in the recent 'windows' outputs from this PR in travis-ci. One issue I find especially vexing is that `utils/tests.sh` has issues using a `#!/usr/bin/env bash` line.

If you wanted to take a crack at making a branch work on travis from your machine like you say, that would be great, it's been surprisingly difficult to get travis-ci windows builds this far. You could start either from my branch or master.  I get that the commit trail might not be helpful, but if you look at the overall diff (41 lines with debug output) I think the intent is pretty clear. 

I've been avoiding using a windows machine/vm here so far, thus the numerous commits. Obviously they'll all get squashed before merge.

Once I get the tests to run, I'll look at the `ps` issue, thanks for bringing that up too.

Mohjive:
Looking at the output from Travis I’d say that it has to do with spaces in paths and the mix of `:` and `;` as path separators. But I don’t have any good advice at how to solve it.

joshrabinowitz:
Hm, maybe it is related to the spaces. 

Looking at https://travis-ci.org/sobolevn/git-secret/jobs/500846698, I'm using `;` as separator on windows and `:` on linux, so I'm pretty sure that's not the problem.

It'd be really preferable to get windows travis-ci tests working so we can test your PR like the rest of the features. So if you do have a few cycles to get git-secret's `make test` further along on travis-ci/windows that would be great. We can temporarily skip some tests on windows if it makes sense because of `ps` differences.

Mohjive:
I can probably put some time into this, when I have access to a windows machine again (after the weekend)

joshrabinowitz:
Ok, I reverted the code that was trying to special-case the PATH var on windows; I wasn't progressing.

This PR is pretty close to as little changes as needed for travis-ci and windows, but doesn't work as you can see at https://travis-ci.org/sobolevn/git-secret/jobs/500921592

I don't know what's going on with things not finding `bash`; according to #40 (windows support) this seems like it should just work.

Mohjive:
I made a PR to your branch for this PR

Mohjive:
@joshrabinowitz I made a PR to your path for this PR

joshrabinowitz:
this PR is being closed, #374 has all the changes in this PR along with PATH and `ps` fixes from @Mohjive

PR #371, opened 2019-03-02
Upgrade bats core

joshrabinowitz:
upgrade bats-core to v1.1.0


joshrabinowitz:
This gives nicer failed-test output than v1.0.2 did

PR #374, opened 2019-03-04
Work to get ci/Travis working for windows

Mohjive:
This is WIP to make Travis work for Windows (git for Windows environment)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,14 +39,21 @@ install-test:
 	git clone --depth 1 -b v1.0.2 https://github.com/bats-core/bats-core.git vendor/bats-core; \
 	fi
 
-# TODO: for PATH below, there has to be a better way to choose : or ; as a delimiter at runtime
+# TODO: For PATH below, there has to be a better way to choose (slash/backslash 
+# TODO: and) ':' or ';' as a delimiter at runtime. Current code is a hack.
 .PHONY: test
 test: install-test clean build
 	chmod +x "./utils/tests.sh"; sync; \
 	export SECRET_PROJECT_ROOT="${PWD}"; \
-	if [[ "${GITSECRET_DIST}" == "windows" ]]; then export PATH="${PWD}/vendor/bats-core/bin;${PWD};${PATH}"; \
-	else export PATH="${PWD}/vendor/bats-core/bin:${PWD}:${PATH}"; fi; \
+	if [[ "${GITSECRET_DIST}" == "windows" ]]; then \
+		GITSECRET_WIN_PATHS=$(echo "${PWD}/vendor/bats-core/bin;${PATH}" | sed -e 's/[/]/\\/g'); \
+		export PATH="${GITSECRET_WIN_PATHS};${PATH}"; \
+	else \
+		export PATH="${PWD}/vendor/bats-core/bin:${PWD}:${PATH}"; \
+	fi; \
 	command -v bash; \
+	command -v bats; \
+	command -v env; \
 	./utils/tests.sh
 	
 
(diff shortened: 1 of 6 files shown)


tests/_test_base.bash
```diff
@@ -30,6 +30,12 @@ BEGIN { OFS=":"; FS=":"; }
 # This command is used with absolute homedir set and disabled warnings:
 GPGTEST="$SECRETS_GPG_COMMAND --homedir=$TEST_GPG_HOMEDIR --no-permission-warning --batch"
 
+# Platform stuff:
+if [[ "$GITSECRET_DIST" == "windows" ]]; then
+PS_CMD="ps -l -u";
+else
+PS_CMD="ps -wx -U";
```
joshrabinowitz: can we indent these to be consistent with the rest of the file?

tests/_test_base.bash
```diff
@@ -66,8 +72,8 @@ function test_user_password {
 
 function stop_gpg_agent {
   local username=$(id -u -n)
-  ps -wx -U "$username" | gawk \
-    '/gpg-agent --homedir/ { if ( $0 !~ "awk" ) { system("kill -9 "$1) } }' \
+  ${PS_CMD} "$username" | gawk \
+    '/gpg-agent --homedir/ { if ( $0 !~ "awk" ) { system("echo -9 "$1) } }' \
```
joshrabinowitz: why this change from 'kill' to 'echo' ?

Makefile
```diff
@@ -42,8 +42,8 @@ install-test:
 .PHONY: test
 test: install-test clean build
 	@chmod +x "./utils/tests.sh"; sync; \
-	export SECRET_PROJECT_ROOT="${PWD}"; \
-	export PATH="${PWD}/vendor/bats-core/bin:${PWD}:${PATH}"; \
+	export SECRET_PROJECT_ROOT="$(shell echo $${PWD})"; \
+	export PATH="$(shell echo $${PWD})/vendor/bats-core/bin:$(shell echo $${PWD}):$(shell echo $${PATH})"; \
```
joshrabinowitz: can we add a comment to explain why we're going through these gyrations with PWD?

joshrabinowitz:
Thank you @Mohjive ! Taking a look

Makefile
Mohjive: Added comments, see commit below

tests/_test_base.bash
Mohjive: Sorry, some lingering debug code. Fixed below

tests/_test_base.bash
Mohjive: Of course, see below.

Makefile
joshrabinowitz: Thank you!

Can we put this comment above the Makefile `.PHONY:test` line so it's not part of the test: command output?

Also maybe a comment like "the shell echo $${PWD}" construct is to get the Unix PATH env var on windows" would be clearer.

joshrabinowitz:
I was reading about msys2 and came across references to 'MSYS2_PATH_TYPE=inherit' which I read as making the msys2 PATH match the windows one.

See "Get Windows Environment to msys2" at https://juniway.github.io/2016/08/26/linux-msys2/ - I couldn't find better docs about this.

I seem to recall some other method of synchronizing env vars (PATH and others) between windows and msys2 but I can't seem to dig up references now.

Mohjive:
I'll look into that.

I did find this when looking into not mangling paths the other day: https://stackoverflow.com/a/34386471, but none of those were working for this issue.

Mohjive:
Seems that **MSYS2_PATH_TYPE=inherit** is used to control wether windows PATH is exposed into MSYS environment, but not how paths are handled.

Mohjive:
Closing this with reference to #375. All work is being done there.

Issue #252, opened 2018-08-31
How to install on Windows?


fede91it:
Hi

How I should install git-secret on Windows OS?

sobolevn:
Hi, @fede91it! Thanks for bringing this up.

I am not a windows user, but there were reports about `git-secret` working on windows some time ago:
- https://github.com/sobolevn/git-secret/issues/40
- https://github.com/sobolevn/git-secret/issues?utf8=%E2%9C%93&q=is%3Aissue+windows

I am pretty sure that `git-secret` will work on ubuntu sub-system.
I would also recommend to try the "Manual" installation method from here: http://git-secret.io/installation

Things you can help us with:
0. provide more OS information (version, service pack, sub-system info, etc)
1. report back if any of these methods worked for you or not
2. super-duper-feature: provide docs for windows folks who will suffer the same problem

Thanks for using `git-secret`. I will appreciate any feedback from you.

joshrabinowitz:
@fede91it: as @sobolevn says, we've had reports about git-secret working well with windows 10 using the "Windows Subsystem for Linux" and Ubuntu <https://docs.microsoft.com/en-us/windows/wsl/install-win10>. 

If you detail here how git-secret does/doesn't work for you on windows I'll add documentation about it to git-secret (unless you beat me to it!). Thanks for filing this issue.

joshrabinowitz:
I was just able to get git-secret's self-tests to pass after cloning the git-secret repo on the Ubuntu system on Windoes 10.

fede91it:
Thank you!

PR #375, opened 2019-03-04
Fix for windows

Mohjive:
This is a merge of work done in #369 and #374


joshrabinowitz:
Right on, @Mohjive ! Excellent progress!

joshrabinowitz:
Awesome! Impressive, @Mohjive . 

I wonder how this branch effects manual installation with WSL on windows 10 -- 
Do the self-tests still pass on windows 10 using WSL as described here? #252

Mohjive:
Yes, "make test" passes for this branch using wsl (I used OpenSUSE Leap). I believe it is equivalent to Linux and as long as the Linux tests pass, wsl should do as well.

joshrabinowitz:
@Mohjive in your opinion is this ready for review and/or merge?

Mohjive:
@joshrabinowitz yes

src/_utils/_git_secret_tools.sh
```diff
@@ -147,6 +147,12 @@ function _os_based {
 
 # File System:
 
+function _clean_path {
+  # This function transforms windows paths to *nix paths
```
joshrabinowitz: Can you show an example of a path transformation in a comment here? For example, 

`# this code would transform c:\this\that.file (or whatever) to XYZ`

joshrabinowitz: Also, how about we rename this function to `_clean_windows_path`, 
since it only effects paths with drive letters (which are a windows thing).
 
If you have a better idea for a function name that works too.

joshrabinowitz:
Also can you please add to CHANGELOG.md as appropriate?
[I can do it in a following PR if you prefer.]

joshrabinowitz:
This PR Closes #375

joshrabinowitz:
I'll add to CHANGELOG.md, rename the function, and add the comment soon.

joshrabinowitz:
This was merged and unmerged because of #409

Issue #376, opened 2019-03-04
Avoid using `ps -wc -U $username | gawk  /.../ { if (...) { system("kill $1") } }` in tests


joshrabinowitz:
Can we do something safer to stop the gpg-agent ? 

Currently in the tests, we do this (excerpted):

````
...
ps -wx -U "$username" | gawk '/gpg-agent --homedir/ { if ( $0 !~ "awk" ) { system("kill -9 "$1) } }' 
...
````
(See https://github.com/sobolevn/git-secret/blob/505cb1fe427cc3dd3f508876679c26d013c54f1d/tests/_test_base.bash#L66 )

Apparently recent gpgs have a command for this:
`gpgconf --kill gpg-agent`

Some suggest doing something like this:
`gpg-connect-agent reloadagent /bye`

Any gpg experts have an opinion on the 'right thing' to do here or is everyone happy with the `kill -9`?

joshrabinowitz:
This comment from closed PR  https://github.com/sobolevn/git-secret/pull/378#issuecomment-469434337  notes:

gnupg (GnuPG) 2.0.22, which is currently on CentOS7/RHEL7, has man pages for gpgconf  that say the `--kill gpg-agent` option is supported, but the gpgconf binaries disagree:

````
% gpgconf --kill gpg-agent
gpgconf: invalid option "--kill"
````

joshrabinowitz:
I'm planning to merge this PR which replaces `kill -9 $pid` with a regular `kill`, which is moderately safer. 

I would like avoid using `kill` altogether but I don't see one best way to do it (see above). Perhaps we should use `pkill`, which would give us clearer and finer-grained control of which process we `kill`.

joshrabinowitz:
Ok, so we're `kill` in place of `kill -9`, which is an improvement.

I'd like to see us move towards either 
a) knowing the correct pid, ideally from something writing a pid file, or 
b) using pkill with an appropriate series of criteria. This is probably clearer and easier to state what we mean than using `ps+gawk+regex+if+system+kill`

joshrabinowitz:
See also #406

PR #378, opened 2019-03-04
use gpgconf to stop gpg-agent, not kill -9

joshrabinowitz:
For #376


joshrabinowitz:
gnupg (GnuPG) 2.0.22, which is currently on CentOS7/RHEL7, has man pages for gpgconf  that say the `--kill gpg-agent` option is supported, but the gpgconf binaries disagree:

````
% gpgconf --kill gpg-agent
gpgconf: invalid option "--kill"
````

PR #379, opened 2019-03-04
use 'kill', not 'kill -9'

joshrabinowitz:
For #376


sobolevn:
@joshrabinowitz does this close #376 completely?

joshrabinowitz:
@sobolevn I don't think so. I'd like to find a cleaner way to stop gpg-agent

joshrabinowitz:
But I think merging this is a step forward, in my tests a regular 'kill' will stop the user's gpg-agent.

tests/_test_base.bash
```diff
@@ -67,7 +67,7 @@ function stop_gpg_agent {
   local username
   username=$(id -u -n)
   ps -wx -U "$username" | gawk \
-    '/gpg-agent --homedir/ { if ( $0 !~ "awk" ) { system("kill -9 "$1) } }' \
+    '/gpg-agent --homedir/ { if ( $0 !~ "awk" ) { system("kill "$1) } }' \
```
joshrabinowitz: This quoting [especially the `system("kill "$1)` part] irks me

Mohjive:
`gpg-connect-agent /bye` (https://superuser.com/a/1075527 and also https://linux.die.net/man/1/gpg-connect-agent)
or
`gpgconf --kill gpg-agent` (https://superuser.com/a/1150399)

joshrabinowitz:
@Mohjive is gpg-connect-agent available in all versions of gpg that we might encounter with git-secret? 

As I mention in #376 , "gnupg (GnuPG) 2.0.22, which is currently on CentOS7/RHEL7, has man pages for gpgconf that say the --kill gpg-agent option is supported, but the gpgconf binaries disagree:"

PR #393, opened 2019-03-06
Verbose env var

### LINKED ISSUE #323: Feature: consider using SECRETS_VERBOSE env var
In PR #321 we wound up discussing if we should add a GIT_SECRET_VERBOSE variable for git-secret-tell. (Edit: to be consistent with existing git-secret env vars, this should probably begin with 'SECRETS_').

@sobolevn said:
>Maybe we can make -v a global option? Something like:
> git secret -v tell user@email.com
> or
> GIT_SECRET_VERBOSE=1 git secret tell some@random.name 

~~We could, but this would be the first time we've had an environment variable modify git-secret's behavior (right?)~~ Edit: We use env vars in git-secret already for SECRETS_GPG_COMMAND, SECRETS_EXTENSION, and SECRETS_DIR; see comment with link to git-secret.7.ronn below.

Maybe verbosity /is/ a special case and we should allow setting it through an env var ~~GIT_SECRET_VERBOSE~~ SECRETS_VERBOSE as well as through the -v option? 

~My immediate gut reaction is to reject is as orthogonal to the way git-secret accepts options (edit: usually) but maybe it is enough of a special case. If you see an important use case for it I'm cool with it.~ 

Edit: I've come to see how it's useful for debugging, testing, and user comprehension.

### LINKED ISSUE #394: don't use gpg --quiet when in decrypting files in verbose mode
When a user is running in verbose mode, we might as well show the output of the gpg decryption command.

Implemented in #393 -- (See: https://github.com/sobolevn/git-secret/pull/393/commits/130f47d300ca66a1e3317fb37dd47fcf2d641c98 )

joshrabinowitz:
Closes #323 
Closes #394

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Upgrade bats-core to v1.1.0, import bats-core into vendor/bats-core (#377)
 - Spelling fixes
 - Fix re: SECRETS_DIR in 'init' and SECRETS_EXTENSION in test_reveal.bats (#364)
+- Use SECRETS_GPG_COMMAND env var in gpg version check (#389)
 - Use Shellcheck on tests/ files, changes for Shellcheck in tests/ (#368)
 
 ## Version 0.2.5
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,12 +3,13 @@
 ## {{Next Version}}
 
 - Support SECRETS_VERBOSE env var to enable verbosity (#323)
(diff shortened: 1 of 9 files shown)


joshrabinowitz:
@sobolevn I think this is about ready to merge after some documentation additions. 

Any other verbosity-related changes (like #395) I'd like to put in a separate PR.

Let me know if you have any feedback.

joshrabinowitz:
Yeah I think this is ready to merge.

PR #400, opened 2019-03-08
use pgrep & kill $pid, not ps+gawk+regex+system+kill

### LINKED ISSUE #376: Avoid using `ps -wc -U $username | gawk  /.../ { if (...) { system("kill $1") } }` in tests

### LINKED ISSUE #403: run shellcheck on osx travis tests

joshrabinowitz:
Closes #376 
Closes #403

### DIFF
--- a/tests/_test_base.bash
+++ b/tests/_test_base.bash
@@ -69,9 +69,9 @@ function stop_gpg_agent {
   local pid
   pid=$(pgrep -U "$username" -x gpg-agent)
 
+  # shellcheck disable=SC2001
   echo "$pid" | sed "s/^/# '$BATS_TEST_DESCRIPTION' gpg-agent pid: /" >&3
 
-  # shellcheck disable=SC2001
   if [[ -n "$pid" ]] && [ "$pid" -gt 0 ]; then
 
     #diagnostic output
--- a/tests/_test_base.bash
+++ b/tests/_test_base.bash
@@ -67,7 +67,7 @@ function stop_gpg_agent {
   local username
   username=$(id -u -n)
   local pid
-  pid=$(pgrep -U "$username" -f "gpg-agent --homedir.*$TEST_GPG_HOMEDIR" )
+  pid=$(pgrep -U "$username" -f "gpg-agent --homedir.*$TEST_GPG_HOMEDIR")
   if [[ -n "$pid" ]]; then
 
     # shellcheck disable=SC2001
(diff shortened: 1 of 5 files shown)


tests/_test_base.bash
```diff
@@ -66,9 +66,13 @@ function test_user_password {
 function stop_gpg_agent {
   local username
   username=$(id -u -n)
-  ps -wx -U "$username" | gawk \
-    '/gpg-agent --homedir/ { if ( $0 !~ "awk" ) { system("kill "$1) } }' \
-    > /dev/null 2>&1
+  local pid
+  pid=$(pgrep -U "$username" -x gpg-agent)
```
joshrabinowitz: Is it important that I preserve the `--homedir` portion of the regex?

Mohjive: Without it `make test` may kill **gpg-agent** for the current user.

joshrabinowitz: That is true. Technically it could still kill the wrong gpg-agent(s), if the user had used `--homedir`. Let me see if I can't improve things a little with `pgrep -f`, which matches against the whole command line

joshrabinowitz: Here's what I'm using now, I think it's pretty much guaranteed not to match the user's `gpg-agent`: 

`pgrep -U "$username" -f "gpg-agent --homedir.*$TEST_GPG_HOMEDIR"`

joshrabinowitz: let me know if you have any thoughts @Mohjive

joshrabinowitz:
I apologize, I merged this into master and the 'fix for windows' branch a little bit prematurely. 

It does not kill the same exact pids as the previous code.

I'm investigating why, and working up a fix if needed.

PR #402, opened 2019-03-08
Diagnose what we're killing in current gpg-agent killing code

joshrabinowitz:
Not intended for merge, for diagnostic purposes


Mohjive:
The issues I had killing and re-launching the agent for Travis using WIndows got me thinking about the reason it's done. My guess is to have a clean key setup for every test, and not testing if launching and killing the agent works. Are we sure.we get a clean environment? Doesn't the agent cache imported keys on file (I know too little about gpg-agent at this moment)? If a clean environment is the goal, wouldn't it be better to launch the agent once, before all tests, and purge keys between the tests?

joshrabinowitz:
@Mohjive I don't know the answer to your question, but it's might be a discussion worthy of its own issue. Feel free to create a github issue on this project about this topic if you want, it's a better place to record the dicussion than here.

PR #413, opened 2019-03-12
Separate test dir

### LINKED ISSUE #407: Do not `rm -rf  /tmp/*`
In function *set_state_initial* there's a line
https://github.com/sobolevn/git-secret/blob/a2a9bd32ba4c7d9c8e14f3c29fe5978a2ea7a99e/tests/_test_base.bash#L201 which is problematic. `${BATS_TMPDIR}` is set to either **/tmp** ot what `${TMPDIR}` is set to when `bats` is run. Neither should be OK to do `rm -rf` on on a normal system.
Tests should either explicitly state what files to delete or use a subdir and delete that.

Related to #360

Mohjive:
This change will set a separate directory for all testing, so that all temporary files are contained and can be cleaned up afterwards. Closes #407

### DIFF
--- a/tests/_test_base.bash
+++ b/tests/_test_base.bash
@@ -26,9 +26,6 @@ BEGIN { OFS=":"; FS=":"; }
 # GPG-based stuff:
 : "${SECRETS_GPG_COMMAND:="gpg"}"
 
-# This command is used with absolute homedir set and disabled warnings:
-GPGTEST="$SECRETS_GPG_COMMAND --homedir=$TEST_GPG_HOMEDIR --no-permission-warning --batch"
-
 # Personal data:
 
 # these two are 'normal' keys
@@ -192,10 +189,13 @@ function remove_git_repository {
 # Git Secret:
 
 function set_state_initial {
-  # Define needed directories
+  # Define needed variables
   TEST_BASE_DIR="${BATS_TMPDIR}/${BATS_TEST_NAME}"
   TEST_RUN_DIR="${TEST_BASE_DIR}/run"
   TEST_GPG_HOMEDIR="${TEST_BASE_DIR}/gpg"
+
+  # This command is used with absolute homedir set and disabled warnings:
+  GPGTEST="$SECRETS_GPG_COMMAND --homedir=$TEST_GPG_HOMEDIR --no-permission-warning --batch"
 
   # Make sure we start clean
   rm -rf "${TEST_BASE_DIR:?}"
--- a/tests/_test_base.bash
+++ b/tests/_test_base.bash
@@ -201,7 +201,7 @@ function remove_git_repository {
 function set_state_initial {
   # Create needed directories
   mkdir -p "${TEST_RUN_DIR}" "${TEST_GPG_HOMEDIR}"
-  cd "${TEST_RUN_DIR]}" || exit 1
+  cd "${TEST_RUN_DIR}" || exit 1
   rm -rf "${TEST_RUN_DIR:?}/*"
 }
 
--- a/tests/_test_base.bash
+++ b/tests/_test_base.bash
@@ -193,7 +193,7 @@ function git_commit {
 
 
 function remove_git_repository {
-  rm -rf ".git"
+  rm -rf "${TEST_RUN_DIR:?}/.git"
 }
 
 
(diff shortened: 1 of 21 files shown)


joshrabinowitz:
@Mohjive I merged in the current master so I could test with the current code base and to prepare for merge of this branch.

utils/tests.sh
```diff
@@ -1,11 +1,23 @@
...
 
-set -e
+function cleanup {
+    rm -rf "${TEST_RUN_DIR}" "${TEST_TMP_DIR}"
+    exit 0
+}
+
+trap cleanup EXIT
 
 # Running all the bats-tests in a dir with spaces:
```
joshrabinowitz: We're introducing and using a bunch of env vars here. 
Can we explain them for the reader/developer?

tests/_test_base.bash
```diff
@@ -11,7 +11,10 @@ source "$SECRET_PROJECT_ROOT/src/_utils/_git_secret_tools.sh"
 # Constants:
 FIXTURES_DIR="$BATS_TEST_DIRNAME/fixtures"
```
joshrabinowitz: Can we add comment(s) here describing where we expect the /run and /gpg directories to be created? 
This will help users understand effect.

Also can the comment describe BATS_TMPDIR and SECRET_PROJECT_ROOT, and how/where it expects those variables to get set?

utils/tests.sh
```diff
@@ -1,11 +1,23 @@
 #!/usr/bin/env bash
 
+set -e
 # `SECRET_PROJECT_ROOT` must be set before running the script.
+[ -z "${SECRET_PROJECT_ROOT}" ] && exit 1
 
-set -e
+function cleanup {
```
joshrabinowitz: I see these directories are created in tests/_test_base.bash. Can we rm them from there too? 

It's cleanest/clearest to have the same block of code be responsible for allocation and de-allocation (creation/destruction) and not another file of code.

tests/_test_base.bash
Mohjive: Sure, I can add those. I'm thinking that maybe I should look at the gpg-agent issue before I submit the final solution to the directory issue. The problem I see is that launching and killing the gpg-agent, unless we can guarantee that we kill the right agent, is bad. Therefore we need a persistent directory to act as homedir. Maybe we should move the requirement of having a running gpg-agent to outside of each separate test?

utils/tests.sh
Mohjive: These are not the same directories. I made an error and used the same names. I can fix that. The TEST_TMP_DIR is for BATS to use as root tmp dir (each test make a subdir in that dir in _test_base_sh). The TEST_RUN_DIR is so that we can use full path when `rm -rf`

tests/_test_base.bash
joshrabinowitz: @Mohjive I think it's fine to merge this branch (when we're happy with it) before fixing the `gpg-agent` issue, which only effects `make test` and has been in the code base for quite some time.

But I agree, we should resolve the `gpg-agent` issue soon.

utils/tests.sh
joshrabinowitz: ah, ok. Maybe some clarifiying comments are in order here too.

tests/_test_base.bash
Mohjive: These changes also only affect `make test` afaik. 

The reason I thought about fixing the gpg-agent issue before I could make this PR final is because I'm unsure how to solve the persistent gpg homedir (if there should be one), and deciding whether there should be a persistent homedir or not decides how and which directories should be created imo.

joshrabinowitz: I'd say we should move the tests towards using a clean environment/new homedir for each test. It's the easiest to get right and the most similar to the current code base. 

I'm not against the persistent homedir idea, but I suspect it's a bigger PR, and I think it might be harder to make logically correct with all the versions of gpg out in the wild.

I'm happy to let you decide when your PR should get merged, too :]

joshrabinowitz:
Hello @Mohjive 

I've been looking into the 'use temp directories for test purposes' issue and have made progress at #442, but that PR doesn't go as far as this one creating test directories (it just tries to make sure we don't write files directly in /tmp).

I welcome your input on #442.  

Also I like the changes this PR makes to `set_state_initial` https://github.com/sobolevn/git-secret/pull/413/files#diff-1df762700e784c99583a24042cc866b9L195 but it seems they're what's causing test failures on some platforms. Any theories what's wrong there?

joshrabinowitz:
Hey @Mohjive I'm going close this because #442 deal with the same issue.  
Thanks for taking a whack at this and the 'ps / kill' topic we discussed.

Issue #386, opened 2019-03-05
address lintian issues found in debian testing of 0.2.3


joshrabinowitz:
We should make sure we've fixed all the lintian issues with manpages mentioned here:https://lintian.debian.org/full/czchen@debian.org.html#git-secret_0.2.3-1

As per #312 'ge tgit-secret updated in debian'

joshrabinowitz:
I believe the only issue flagged there that might still be an issue is:

````
W manpage-has-bad-whatis-entry
    usr/share/man/man7/git-secret.7.gz
````

Looks like this is because the NAME section on git-secret.7.ronn is not as expected

https://lintian.debian.org/tags/manpage-has-bad-whatis-entry.html says about this:

Each manual page should start with a "NAME" section, which lists the name and a brief description of the page separated by "\-". The "NAME" section is parsed by lexgrog and used to generate a database that's queried by commands like apropos and whatis. This tag indicates that lexgrog was unable to parse the NAME section of this manual page.

For manual pages that document multiple programs, functions, files, or other things, the part before "\-" should list each separated by a comma and a space. Each thing listed must not contain spaces; a man page for a two-part command like "fs listacl" must use something like "fs_listacl" in the "NAME" section so that it can be parsed by lexgrog.

Refer to the lexgrog(1) manual page, the groff_man(7) manual page, and the groff_mdoc(7) manual page for details.

Severity: normal, Certainty: certain

Check: manpages, Type: binary

joshrabinowitz:
See Also #437

PR #422, opened 2019-03-15
show full commands, including args, in Makefile

joshrabinowitz:
For #386 

As per recommendations at
https://wiki.debian.org/UpstreamGuide

````
Verbose builds

We need to record as much detail from the build log as possible so that we can debug 
build failures and other strange issues. Your build system may have a silent building 
option to hide the full commands (including arguments) used during building. If it 
does, we recommend that it be off by default or that one of the following commands 
enables verbose building.

    ./configure --disable-silent-rules
    make V=1
````

Edit: We don't use ./configure, and I don't think we want to support `V=1` as shown, so I recommend showing the commands run from the Makefile (IE, removing the preceding `@` symbols from commands)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@
 
 - Use Shellcheck on tests/ files, changes for Shellcheck in tests/ (#368)
 - Use Shellcheck on MacOS/osx travis tests (#403)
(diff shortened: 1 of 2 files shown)


joshrabinowitz:
Planning to merge this tomorrow if there's no resistance.

Issue #431, opened 2019-03-20
Tests are sensitive to contents of test's parent directories


joshrabinowitz:
If any parent directory of the directory used to perform the `git-secret` tests contains `.git`, the tests will unexpectedly fail.

joshrabinowitz:
You can see the problem here: https://travis-ci.org/sobolevn/git-secret/jobs/508983357

* All 92 tests pass at first: 
  https://travis-ci.org/sobolevn/git-secret/jobs/508983357#L806

* But after initializing a git repo in /tmp/ with `(cd /tmp; git init)`, we get three test failures starting just below here:
  https://travis-ci.org/sobolevn/git-secret/jobs/508983357#L849

PR #432, opened 2019-03-20
Test will fail if parent dir of test directory contains .git

joshrabinowitz:
This PR exhibits the issue described in #431


joshrabinowitz:
This branch is not intended to be merged, it is to exhibit issue #431 . Therefore closing. 
This is still relevant to #431

Issue #412, opened 2019-03-11
New feature : sops support for encryption


learsix:
Hi,

I'm using [sops](https://github.com/mozilla/sops) in several projects and I've searched how to integrate it with git.

For the moment I'm only using gpg encryption with sops so it made sense to try to adapt git-secret. I've been doing some work in my own fork (no test or doc yet).

Would you be interested to merge this feature ?

joshrabinowitz:
I think it would be interesting to see the PR you've worked up.  I hadn't heard of SOPS until just now.

I can't speak to the likelihood of the PR being merged/accepted (especially as git-secret is relatively easy to describe now, being built on gpg), but it sounds like an interesting project.

Thanks for making us aware! I look forward to seeing your PR (though again I cannot guarantee acceptance of your PR)

joshrabinowitz:
@learsix I've been thinking about this more, and how it would be nice to add SOPS support to git-secret in theory.  

Could it be implemented by adding a `-S`switch to `git-secret`? What would a first integration of `git-secret` and `SOPS` https://github.com/mozilla/sops look like from a user's perspective?

learsix:
@joshrabinowitz I'm almost through a first implementation. I still have tests to add before submitting a PR (I try to stick to the contributing rules ;) I'm not sure I'll have time to push it by the end of the week but it will definitely be done in the next days.

Below a few indications about the way I've implemented it. If you have feedback, I'd be happy to update the code before pushing the PR.

### Impacts for user / git-secrets commands

From the user's perspective it's integrated as described below. All other commands are not impacted. Everything defaults to git-secret as it works now.

**choosing btw `sops` and `pgp`**

`sops` or legacy `pgp` "mode" is chosen at the repo level. The user can define it with the init command and it defaults to `pgp`.

Example: `git secret init -m sops`

Notes:
* based on your previous comment, maybe you'd like to replace `-m sops` with a `-s`
* for now, mode can be changed by re-applying `git secret init` with a different `-m` option. I've not changed messages so the user won't guess it. There will still be work to do here.

**using sops group feature**

To support groups (ie. Shamir Secret Sharing), the user can choose a group when an email address is included. If no group is provided it defaults to a `default` group.

Example:  `git tell -g group1 foo.bar@example.org`

To display groups I've extended the `whoknows` command with the long display option to add `-group <groupname>` at the end of each line. It's only displayed when `sops` mode is enabled in the repo.

**Using keyservice during decryption**

[...]

PR #435, opened 2019-03-20
sops support for encryption

learsix:
This PR implements sops as an alternative to gpg for file encryption / decryption.

Related issue #412 

N/A

Any other comments?
-------------------
* All command options should be supported with `sops` (I mentioned in  #412 that some may not be supported for sops, e.g. `-p`).
* tests should probably be enhanced for sops. Group decryption and keyservice usage are not covered yet
* in main.sh, we detect whether the script is called as git-secret or as gpg command wrapper for sops. This is not clean (as it depends on gpg params passed by sops), I would suggest to use softlinks to sops script instead and detect how the script is used base on `$0`


joshrabinowitz:
It looks like this is failing (at least in part) because sops is not installed on the travis test VMs. 

Perhaps add code to clone and install sops for the travis environments? 
Let me know if you'd like me to give it a try (can't guarantee I'll get to it).

joshrabinowitz:
@learsix I was looking at the code and realized that the tests seem to be using fd 3 for two purposes: 

 - bats-core outputs to fd 3 for 'diagnostic' messages
 - it looks like the new `_gpg_sops_wrapper_encrypt` uses fd 3 to for the passphrase.

learsix:
@joshrabinowitz CI did indeed failed because sops was not installed : tests should be skipped if sops is not installed. I'm correcting this. Then I'll figure out how to add sops with travis ; I think we need to make sure that git-secret still works with gpg only and add sops as an option.
You're right about fd 3, I'm not sure that it impacts tests (`make test` is working locally for me). I'll look into it.

learsix:
@joshrabinowitz fd3 will not impact tests. Using sops, git-secret forks to launch sops and sops forks to launch gpg wrapper (ie. git-secret / _gpg_sops_wrapper_encrypt). At this stage even if fd3 is inherited we overwrite it with a new fd.

learsix:
Last commit should pass CI tests - without sops. Next thing to do is adding sops in CI env to actually test it.

joshrabinowitz:
@learsix are you happy with this PR as it stands?

joshrabinowitz:
@learsix I'm going to close this PR so it's not floating around being open. Feel free to re-open it and continue development if you get the urge! Thank you.

learsix:
@joshrabinowitz Sorry I've been too busy in the past few weeks to finish this. I've started to test my fork in the field. They are still a few updates / tests to add. I'll let you know when it's done.

PR #437, opened 2019-03-21
respect DESTDIR as per GNU/debian/etc

### LINKED ISSUE #424: use DESTDIR in Makefile for debian
This is part of #386 

#### The TL;DR:

We should probably be installing into $DESTDIR/$PREFIX

#### Details:

According to the Debian Upstream Guide https://wiki.debian.org/UpstreamGuide we should be using DESTDIR (they don't mention PREFIX)

> Installation
> Please honor the DESTDIR environment variable or make variable and use its contents as a prefix in front of all installation paths. ...

Gnu coding standards agree at https://www.gnu.org/prep/standards/html_node/DESTDIR.html

>DESTDIR is a variable prepended to each installed target file, like this:
>
>   $(INSTALL_PROGRAM) foo $(DESTDIR)$(bindir)/foo
>   $(INSTALL_DATA) libfoo.a $(DESTDIR)$(libdir)/libfoo.a
>
>The DESTDIR variable is specified by the user on the make command line as an absolute file 
>name. For example:

>   make DESTDIR=/tmp/stage install

FreeBSD docs at https://www.freebsd.org/doc/en_US.ISO8859-1/books/porters-handbook/porting-prefix.html give even more details:

> PREFIX determines where the port will be installed. It defaults to /usr/local, but can be set by the user to a custom path like /opt. The port must respect the value of this variable.
>
> DESTDIR, if set by the user, determines the complete alternative environment, usually a jail or an installed system mounted somewhere other than /.

joshrabinowitz:
Closes #424

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 - Add -v 'verbose' option to 'tell', showing gpg output (#320)
 - Change 'init' to never ignore .secret files (#362)
 - 'add' appends filepaths to .gitignore by default (#225)
+- Automate the GitHub release (#411)
 
 ### Bugfixes
 
@@ -25,7 +26,7 @@
 - Fix re: SECRETS_DIR in 'init' and SECRETS_EXTENSION in test_reveal.bats (#364)
 - Use SECRETS_GPG_COMMAND env var in gpg version check (#389)
 - Add header to git-secret.7 man page, for debian and doc improvement (#386)
-- Automate the GitHub release (#411)
+- Respect DESTDIR when installing as per GNU/debian/etc recommendations (#424)
 
 ## Misc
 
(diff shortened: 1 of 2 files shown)


joshrabinowitz:
With this PR, running as a non-privileged user:

````
% make install
chmod +x "./utils/install.sh"; sync; \
"./utils/install.sh" ""/usr""
cp: cannot create regular file ‘/usr/bin/git-secret’: Permission denied
make: *** [install] Error 1

% DESTDIR=/tmp PREFIX=/usr/local make install
chmod +x "./utils/install.sh"; sync; \
"./utils/install.sh" "/tmp/usr/local"
Installed git-secret to /tmp/usr/local/bin/git-secret

% ls /tmp/usr/local/bin/git-secret
/tmp/usr/local/bin/git-secret*
````

PR #430, opened 2019-03-20
Better temp dir and output

joshrabinowitz:
Note
*  new use of `git check-ignore -q `, 
* removal of directory creation/destruction in  `_secrets_dir_is_not_ignored`

Also relates to #407 and #360 as it sets TMPDIR (and thus stops BATS_TMPDIR from using the default /tmp).

PR #440, opened 2019-03-23
Use git check ignore

joshrabinowitz:
Splitting #430 into multiple PRs. Part 1 of 3.

 This uses `git check-ignore` to test if files are ignored by git instead of `git add`.
Also removes unneeded create/delete directory code.


joshrabinowitz:
This looks good, planning to squash-merge this tomorrow.

PR #442, opened 2019-03-23
Use /tmp/subdir for testing, not /tmp.

joshrabinowitz:
Closes: #407 

This is #430 split into multiple PRs. Part 3 of 3.

This PR changes the test to happen in a subdir of /tmp, which apparently causes 
https://github.com/sobolevn/git-secret/pull/430#issuecomment-475865438 (travis very occasionally fails tests on ubuntu)

PR #443, opened 2019-03-23
Verbose testing

joshrabinowitz:
Adds 'verbose' output from test code to #442 for developer testing


joshrabinowitz:
Closed without merging. In #445 we capture output from subcommands into a temp file, which is more complex but allows us to choose whether to display the captured output or not at runtime.

PR #457, opened 2019-04-01
Tmp cleanups

joshrabinowitz:
We only show 'cleaning up /tmp/file' messages if we're in a VERBOSE mode.


joshrabinowitz:
@hurricanehrndz it looks like you originally wrote the code messaging the deletion of the tmp file.

Let me know if you have resistance to merging this PR!

PR #462, opened 2019-04-27
replace grep+cut+sed pipe with awk regex

simbo1905:
Replaces piping through grep, cut and sed with a single awk regex. This improves efficiency and simplifies the code. 

N/A

N/A

Any other comments?
-------------------
Finally found some time to implement some larger features and noticed this while stepping through the code using bashdb. Thought to send it as a small PR rather than to mix it into a larger feature.

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -628,7 +628,7 @@ function _get_users_in_gpg_keyring {
   # pluck out 'uid' lines, fetch 10th field, extract part in <> if it exists (else leave alone).
   # we use --fixed-list-mode so older versions of gpg emit 'uid:' lines.
   # gensub at the end is to extract email from <>. (If there's no <>, then the line is just an email address anyway.)
-  result=$($SECRETS_GPG_COMMAND "${args[@]}" --no-permission-warning --list-public-keys --with-colon --fixed-list-mode | awk -F: '$1~/uid/{print gensub(/.*<(.*)>.*/, "\\1", "g", $10); }')
+  result=$($SECRETS_GPG_COMMAND "${args[@]}" --no-permission-warning --list-public-keys --with-colon --fixed-list-mode | gawk -F: '$1~/uid/{print gensub(/.*<(.*)>.*/, "\\1", "g", $10); }')
 
   echo "$result"
 }
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -627,8 +627,8 @@ function _get_users_in_gpg_keyring {
 
   # pluck out 'uid' lines, fetch 10th field, extract part in <> if it exists (else leave alone).
   # we use --fixed-list-mode so older versions of gpg emit 'uid:' lines.
-  # sed at the end is to extract email from <>. (If there's no <>, then the line is just an email address anyway.)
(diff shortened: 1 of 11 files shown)


src/_utils/_git_secret_tools.sh
```diff
@@ -628,8 +628,8 @@ function _get_users_in_gpg_keyring {
 
   # pluck out 'uid' lines, fetch 10th field, extract part in <> if it exists (else leave alone).
   # we use --fixed-list-mode so older versions of gpg emit 'uid:' lines.
-  # sed at the end is to extract email from <>. (If there's no <>, then the line is just an email address anyway.)
-  result=$($SECRETS_GPG_COMMAND "${args[@]}" --no-permission-warning --list-public-keys --with-colon --fixed-list-mode | grep ^uid: | cut -d: -f10 | sed 's/.*<\(.*\)>.*/\1/')
+  # gensub at the end is to extract email from <>. (If there's no <>, then the line is just an email address anyway.)
+  result=$($SECRETS_GPG_COMMAND "${args[@]}" --no-permission-warning --list-public-keys --with-colon --fixed-list-mode | gawk -F: '$1~/uid/{print gensub(/.*<(.*)>.*/, "\\1", "g", $10); }')
```
joshrabinowitz: @simbo1905 can you please add a comment explaining your use of gawk here a little more? 

As a reviewer I'm not familiar enough with gawk to say much more than that I can follow the code and it appears to work.

simbo1905: I have made a change to the comment. 

A [wise person once told me](https://codereview.stackexchange.com/a/212095/75693) that "In general, any combination of sed, awk, and grep would be better expressed using just an AWK script." So I have been trying to become more familiar with it. Across the git-secret codebase I see examples of using awk and gawk. One is usually symlinked to another. When it comes to doing specific computations using the more specific version seems to make sense so I went with gawk here.

joshrabinowitz: Not sure I agree about always favoring `(g)awk` over `grep/cut/sed`, sure we gain some speed and save some keystrokes, but at what cost to readability? How many languages do developers need to know just to make small changes to git-secret?

That having all been said, I'm good with this change if we think it's logically correct.

joshrabinowitz: This PR does use some gawk constucts I've never seen before too, though. Do we know if we need a particular minimum version of gawk to support this use?

simbo1905: No I am not aware of a minimum version.

joshrabinowitz: I was really implying we should make sure we didn't need _too_ recent a version of `gawk` to support `git-secret`. Also it would be good if we could tell users a minimum required version.

joshrabinowitz: It's not a big enough deal to worry about if you're not inclined though. LGTM.

joshrabinowitz:
This PR LGTM, moduli my _mild_ preferences for figuring out what the minimum version of gawk we'd need to support the features used in this PR.  (Which I presume are more modern than most other gawk features used in `git-secret`). 

There's a gawk change history here if we wanted to try to figure it out that way:
  https://www.gnu.org/software/gawk/manual/html_node/Feature-History.html

But anyway, I'm cool with this getting merged as-is if @simbo1905 desires. Thank you Simon!

simbo1905:
gensub was gawk 3 which as a release dated 2010. the rest of the syntax is regular field pattern selection.

simbo1905:
the readme says we depend up `gawk` since `4.0.2` so when the refresh passes i will merge 🍾

PR #464, opened 2019-04-27
remove dead code for keys/mapping.cfg

simbo1905:
Removes dead code. 

N/A

N/A

Any other comments?
-------------------
I was trying to figure out what keys/mapping.cfg does then realise that i could not find any code that uses it. Deleting it and any references to it doesn't break any test. I feel slightly uncomfortable in deleting the code since I don't yet consider myself familiar with the codebase. I am also unsure whether test coverage is sufficient to really confirm that this is indeed dead code. Was this something that was used in the past but was made redundant by a refactor?


joshrabinowitz:
I don't think use of `_SECRETS_DIR_KEYS_MAPPING` or `.gitsecret/keys/mapping.cfg` was ever implemented, based on my reading of the code and history. 

Looks like they were only ever mentioned in commit 9d0a2ac1c19b5edfc34e6d08e120764b407df92c

LGTM

Issue #451, opened 2019-03-29
Fix case when tester sets TMPDIR


joshrabinowitz:
This is to track the bug we saw in 

-   https://github.com/sobolevn/git-secret/pull/442#issuecomment-475877864

see for example here:

-    https://travis-ci.org/sobolevn/git-secret/jobs/512538180#L1972 

(which is from a different branch, but shows the same issue) when we set TMPDIR to an existent /tmp/tempdir

joshrabinowitz:
This is fixed by #473

Issue #360, opened 2019-02-20
find and rm only the relevant files when a test fails


joshrabinowitz:
As mentioned in #334 , when tests fail users might see a long string of find errors, such as:

````
not ok 28 run 'hide' from inside subdirectory
# (in test file /home/joshr/gitsrc/git-secret/tests/test_hide.bats, line 88)
#   `[ "$status" -eq 0 ]' failed
# rm: cannot remove ‘space file’: No such file or directory
# fatal: Not a git repository: '.git'
# find: ‘/tmp/tmp.aBlVDAbUQF’: Permission denied
# find: ‘/tmp/systemd-private-8b7dddc415874f87ad46f53ff94b4d7f-rtkit-daemon.service-KNfvrc’: Permission denied
# find: ‘/tmp/systemd-private-8b7dddc415874f87ad46f53ff94b4d7f-colord.service-ivZa8u’: Permission denied
# find: ‘/tmp/systemd-private-8b7dddc415874f87ad46f53ff94b4d7f-fwupd.service-07aFP3’: Permission denied
# find: ‘/tmp/tmp.UbZvBTtYAa’: Permission denied
# find: ‘/tmp/systemd-private-8b7dddc415874f87ad46f53ff94b4d7f-chronyd.service-G1ov1r’: Permission denied
# find: ‘/tmp/systemd-private-8b7dddc415874f87ad46f53ff94b4d7f-cups.service-kAx3OS’: Permission denied
# find: ‘/tmp/systemd-private-8b7dddc415874f87ad46f53ff94b4d7f-bolt.service-IZaN9N’: Permission denied
# find: ‘/tmp/tmp.1L9d7x7l3I’: Permission denied
````

This shouldn't happen. See also #333

Mohjive:
It would be easy to add a `2>/dev/null` at the end of the `find | xargs rm` command in **function _find_and_clean** at line ~442 in src/_utils/_git_secret_tools.sh to hide all such warnings. The verbose output would still print. Is this an OK solution?

joshrabinowitz:
No let's not do that. 

If a test fails, we should not trying to 'rm' files in /tmp that had nothing to do with our code

joshrabinowitz:
See Also: #452

joshrabinowitz:
This is fixed in #457.

joshrabinowitz:
Fixed by #473

PR #473, opened 2019-05-15
Tempfile and temp directory cleanups

joshrabinowitz:
For #451, #360, and general improved use of tmpfiles and tmpdirs.


joshrabinowitz:
If there's no resistance, I'll plan on merging this in the next few days.

Issue #487, opened 2019-06-11
fix fedora builds to not rely on python2-dnf and re-enable fedora builds


joshrabinowitz:
Recent versions of fedora no longer include the package `python2-dnf`, 
so the travis builds fail on fedora now. 

See: https://github.com/ansible/ansible/issues/54855

joshrabinowitz:
We get the build-time error:

````
       failed: [localhost] (item=gawk) => { 
    ... "cmd": "dnf install -y python2-dnf",  
    ... "msg": "Error: Unable to find a match", ... 
     ]}
````
See: https://travis-ci.org/sobolevn/git-secret/jobs/544174411#L2586

joshrabinowitz:
Not sure how to fix this one. @simbo1905 / @hurricanehrndz / @sobolevn  ?

sobolevn:
Some googling showed these results:
- https://bugzilla.redhat.com/show_bug.cgi?id=1696404
- https://github.com/ansible/ansible/issues/54855

Maybe we can try to use `python3` and `python3-dnf`?

simbo1905:
Python 2 stops getting regular updates in 2020 so rhel8 and centos8 dont make it the default and instead folks need to choose between python2 or python3. Moving to python 3 seems sensible as its release date was 2008.

joshrabinowitz:
@simbo1905, @sobolevn agreed, we need to move to a modern version of python. 
Either of you up to perform this change?

joshrabinowitz:
more details: Our build system (which uses python) has sort of been grandfathered in and we've been getting warnings in the builds about how we should move off python2 for some time now. 

I'm hoping someone other than me is up to fixing this build issue as I'm mostly concentrated on the code in git-secret proper. Thanks in advance!

joshrabinowitz:
See also: https://docs.ansible.com/ansible/latest/reference_appendices/python_3_support.html

joshrabinowitz:
I think we're hitting neillturner/omnibus-ansible/issues/25 "Support for pip3" in omnibus-ansible, which is used by test-kitchen to install ansible with pip

joshrabinowitz:
Fixed in #511

PR #490, opened 2019-06-13
Add SECRETS_PINENTRY env var and docs

joshrabinowitz:
For #221


joshrabinowitz:
This can't work because --pinentry-more doesn't accept a program argument. See `man gpg` on gpg after 2.1

joshrabinowitz:
retooling to see if we can use SECRETS_PINENTRY to specify 'default', 'ask', 'cancel', 'error' or 'loopback'.

joshrabinowitz:
This is blocked from merging by #487

PR #492, opened 2019-06-25
don't use revoked gpg uid

bricewge:
fix sobolevn#491

I dug a little further than the what I wrote in the bug report and fixed my particular issue. However this patch isn't fail proof since it only exclude revoked uid but there maybe other [validity values](https://github.com/gpg/gnupg/blob/master/doc/DETAILS#field-2---validity) that trigger gpg to try to get keys remotely (WKD, keyserver or whatnot).


joshrabinowitz:
@bricewge Great input, thank you for your PR.

joshrabinowitz:
Blocked by #487

PR #505, opened 2019-08-16
disable tests to get travis-ci passing

joshrabinowitz:
Disable travis tests on Fedora and Debian temporarily


joshrabinowitz:
Debian tests have been restored,
Fedora builds are still temporarily disabled because of blocker #487

Issue #241, opened 2018-07-15
Test that version of gpg used by kitchen matches test name (gnupg1/gnupg2)


joshrabinowitz:
gnupg1-ubuntu-latest and gnupg1-ubuntu-rolling builds actually test with gpg2, not gpg1.

The evidence: search for 'against gpg' in the outputs of the builds at (for example) 
https://travis-ci.org/sobolevn/git-secret/builds/404141903?utm_source=github_status&utm_medium=notification . 

gnupg1-ubuntu-latest and gnupg1-ubuntu-rolling builds show 'Running test against gpg (GnuPG) 2.2.4.'

gnupg1-debian-latest and gnupg1-fedora-latest both use gpg1 (1.4.21) as expected

simbo1905:
This ticket as it may cause confusion if someone tries to fix this issue in isolation of #235.

I agree that we should test against gpg2.

I don’t think we should test against `rolling`. Looking at hub.docker.com currently `rolling` and `latest` are currently the same image. Reading what is stated at about rolling it reads that we can expect it to be stable than latest. I don’t think it’s really in our interests to test against anything less stable than the latest release as the vast majority of our users will be using that or something older. 

On #235 I can upgrade to using gpg2 against the current major release on master/PR and use the latest release and gpg2 for tagged deploy builds. 

Closing this so that it won’t clash with #235.

joshrabinowitz:
I confirmed that only the `gnupg1-ubuntu-*` builds are using the wrong version of GnuPG:

````
% grep -i 'running test against' *txt
gnupg-git-fedora-latest.txt:           "msg": "Running test against gpg (GnuPG) 2.3.0-unknown."
gnupg1-alpine-latest.txt:           "msg": "Running test against gpg (GnuPG) 1.4.23."
gnupg1-debian-latest.txt:           "msg": "Running test against gpg (GnuPG) 1.4.21."
gnupg1-fedora-latest.txt:           "msg": "Running test against gpg (GnuPG) 1.4.23."
gnupg1-ubuntu-latest.txt:           "msg": "Running test against gpg (GnuPG) 2.2.4."
gnupg1-ubuntu-rolling.txt:           "msg": "Running test against gpg (GnuPG) 2.2.4."
````

joshrabinowitz:
We should add a test that confirms that the version of gnupg used matches what the kitchen spec expects. 

[...]

PR #506, opened 2019-08-16
when we say gnupg1 don't use gnupg2

joshrabinowitz:
Use gnupg version 1 in Alpine and Fedora gnupg1 tests


joshrabinowitz:
Relates to #241

PR #523, opened 2019-08-21
Issue 478 tests in busybox

### LINKED ISSUE #478: fix tests when using busybox
There's a few issues we see testing with busybox (such as on Alpine Linux without coreutils installed)

** we get an error at `tests/test_hide.bats` line 86:

` [ " $secret_perm" = "$file_perm" ] failed`
`ps: unrecognized option: w`

There's multiple issues here: umask might make our permissions test fail; busybox's `ps` and `stat` are different than `coreutils`'s

See also #477 which tries to deal with more of these issues.

joshrabinowitz:
Closes #478

### DIFF
--- a/.Dockerfiles/alpine/latest/Dockerfile
+++ b/.Dockerfiles/alpine/latest/Dockerfile
@@ -1,9 +1,9 @@
 FROM alpine:latest
 
+# don't install coreutils, so we get busybox versions of stat and mktemp. See #475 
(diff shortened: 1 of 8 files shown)


joshrabinowitz:
I think this is about ready to merge. Any input @sobolevn ?

sobolevn:
Thanks a lot! @joshrabinowitz awesome work! 👍

Issue #544, opened 2019-09-20
correct version mention in git-secret add man page


joshrabinowitz:
Man pages for git-secret-add say:

"When adding files to encrypt, `git-secret-add` (as of 0.3.0) will ensure that they are ignored by `git` by mentioning
 them in .gitignore,"

This change was actually released in the version 0.2.6 on Jun 9 2019

joshrabinowitz:
Closed by commit shown above

Issue #420, opened 2019-03-15
git-secret.io should mention which version of git-secret it is documenting


joshrabinowitz:
git-secret.io ~documents~ should document the last _released_ version of git-secret, and not the version in `master`. 
Maybe this should be mentioned on git-secret.io.

PR #545, opened 2019-09-20
put version in man pages, update man page dates,  fix version mention

joshrabinowitz:
For #420 and #544


joshrabinowitz:
This PR gets man pages to list the name of git-secret as `git-secret 0.3.1` in the man page headers, like so:

`GIT-SECRET-ADD(1)       git-secret 0.3.1         GIT-SECRET-ADD(1)`

Issue #508, opened 2019-08-17
Skip invalid keys in keyring in 'tell'


joshrabinowitz:
https://github.com/sobolevn/git-secret/issues/491#issuecomment-505492667 and #492 relate to gnupg [validity values](https://github.com/gpg/gnupg/blob/master/doc/DETAILS#field-2---validity).

~We currently skip any 'revoked' keys (see #492) but there are other cases of keys we should probably skip like 'disabled', 'expired', and 'disabled' keys.~ Edit: We no longer skip revoked keys in git-secret

joshrabinowitz:
We should also add tests for other invalid keys we skip (like #510, but for other invalid key types)

joshrabinowitz:
On second thought: If we want git-secret to detect invalid keys, there are several cases of invalid keys, and since gnupg is not particularly helpful in reporting problems with keys, we should probably perform error reporting about why we don't use particular invalid keys.

joshrabinowitz:
See also #238 regarding 'expired' keys

joshrabinowitz:
@sobolevn @simbo1905 thoughts on the issue of having git-secret detect and report on unusable keys?

simbo1905:
I agree that #516 seems the right direction.

joshrabinowitz:
Closed as 'wontfix'

joshrabinowitz:
Reopening, this will resolve #552

joshrabinowitz:
@sobolevn / @simbo1905 , in light of the above comment, I'm planning fix this issue instead of closing as 'wontfix'.

joshrabinowitz:
Closed by #553 , which issues a warning for expired/revoked/invalid keys

sowbug:
Commenting on this closed issue because the behavior seems to be covered by the original issue description.

I'm unable to `git secret tell` an email address with two keys, one of which is disabled. I'm running 0.5.0. Is this expected?

If I `gpg --edit-key` for the disabled key, underneath the primary key it says `*** This key has been disabled`, which according to the gpg manual means "A disabled key cannot normally be used for encryption."

Yet when I `git secret tell` the email address, the response is `git-secret: abort: 2 keys found in gpg user keyring for: [address]`. While that's true, it seems intuitively that git-secret should ignore disabled keys.

Am I misunderstanding the behavior?

Note that I don't want to delete the second key, as it's still used for decryption of old files. It hasn't been revoked, and it isn't expired, either.

Issue #317, opened 2019-01-16
Feature request: better messaging about revoked keys


joshrabinowitz:
See discussion in #283 
This has actually come up in the wild, see #280

joshrabinowitz:
See also #508 and #238

joshrabinowitz:
This has come up again and is handled in multiple PRs including #553

joshrabinowitz:
Closed by #553

Issue #290, opened 2018-12-13
Cannot encrypt for a certain set of keys.


hickscorp:
A colleague of mine has several keys. Once I `git secret tell` him, the cannot be encrypted anymore (`git secret hide` has non-zero exit) but **no error message is shown** which IMO is quite critical because it could be disregarded by a human.

I believe the failure is caused by one of the keys owned by `jean dot parpaillon at free point fr` (He authorised me to mention his address here for conducting further tests).

joshrabinowitz:
Hello @hickscorp ,

Thanks for filing this issue. I looked into this a little more, and according to <https://pgp.surfnet.nl/pks/lookup?search=jean.parpaillon%40free.fr&fingerprint=on&op=index>) the key for that email address has been revoked.

The error checking and logging in git-secret have been improved since the release of 0.2.4.  
So the current master branch will give you better error messages if a file can't be encrypted or decrypted.

joshrabinowitz:
If there's something I'm missing here please let me know

hickscorp:
@joshrabinowitz Thanks for the update. Sorry, just got back home from a trip abroad.
I'm using `git-secret` version `0.2.2` as that's the package that gets installed when running `apt-get install git-secret` on Xubuntu 18.04. Ideally I'd like to stay with mainstream packages - having the binary manually installed is the best way for me to forget to update it ;)

Thanks! I'll keep checking on the `deb` packages for updates.

joshrabinowitz:
@hickscorp we have made changes about error reporting and handling of expired, revoked, and otherwise invalid keys. 

Also you can install current versions of git-secret using deb packages at: 
https://dl.bintray.com/sobolevn/deb/
We are still working on getting git-secret updated in debian

joshrabinowitz:
Closed by #553

PR #553, opened 2019-10-02
Issue 552 508 revoked keys

### LINKED ISSUE #552: git-secret-hide fails when encrypting with a revoked and active key for the same email, so help user avoid this situation
What are the steps to reproduce this issue?
-------------------------------------------
1. Create a new git repository, with a file you'd like to encrypt
2. `git secret tell` a key with multiple identities
3. `git secret hide`

What happens?
-------------
git-secret aborts.

What were you expecting to happen?
---------------------------------
The hide to succeed.

Any logs, error output, etc?
----------------------------
```
$ git secret hide
gpg: error retrieving 'email@example.com' via WKD: No data
gpg: email@example.com: skipped: No data
gpg: /path/to/example/secret.txt: encryption failed: No data
git-secret: abort: problem encrypting file with gpg: exit code 2: secret.txt
```

Any other comments?
-------------------
I verified that another key works just fine.

This is they key I'm trying to `git secret tell` to: [DD2C 4F19 5E70 A0D9 2F86  2D6F 60BA 2A8B FD58 3641](https://keybase.io/cantido/pgp_keys.asc).

There are several revoked identities on that key, and the error message says the error is in retrieving one of the revoked identities. I used one of the valid identities when I added the key via `git secret tell`. All identities, including the revoked ones, are listed by `git secret whoknows`.

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`)
`Darwin hostname 18.7.0 Darwin Kernel Version 18.7.0: Tue Aug 20 16:57:14 PDT 2019; root:xnu-4903.271.2~2/RELEASE_X86_64 x86_64`

**`git-secret` path:** (`which git-secret`)
`/usr/local/bin/git-secret`

**`git-secret` version:** (`git secret --version`)
`0.3.1`

**`git` version:** (`git --version`)
`git version 2.23.0`

**Shell type and version:** (`$SHELL --version`)
`zsh 5.3 (x86_64-apple-darwin18.0)`

**`gpg` version:** (`gpg --version`)
```
gpg (GnuPG) 2.2.17
libgcrypt 1.8.5
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <https://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Home: /Users/user/.gnupg
Supported algorithms:
Pubkey: RSA, ELG, DSA, ECDH, ECDSA, EDDSA
Cipher: IDEA, 3DES, CAST5, BLOWFISH, AES, AES192, AES256, TWOFISH,
        CAMELLIA128, CAMELLIA192, CAMELLIA256
Hash: SHA1, RIPEMD160, SHA256, SHA384, SHA512, SHA224
Compression: Uncompressed, ZIP, ZLIB, BZIP2
```

### LINKED ISSUE #587: `git secret hide` should skip revoked identities
What are the steps to reproduce this issue?
-------------------------------------------
1. Run `git secret tell` with some GPG key that has multiple email addresses, some of which have been revoked.
2. Run `git secret hide` and watch it fail with GPG error messages.

What happens?
-------------
GPG errors out, and so does `git-secret` (sorry, back-translated from italian-language locale, error messages might not be exactly those output by GPG in `en_US`):
```
gpg: riccardo.murri@+++: skipped: public key already added
gpg: riccardo.murri@---: skipped: public key already added
gpg: error retrieving 'murri@***' via WKD: missing data
gpg: murri@***: skipped: missing data
gpg: etc/service.properties: encryption failed: missing data
git-secret: abort: problem encrypting file with gpg: exit code 2: etc/service.properties
```

What were you expecting to happen?
----------------------------------
That revoked GPG identities are skipped.

They are marked with `r` in the second field of GPG's `--with-colon` list output.

What versions of software are you using?
----------------------------------------
**Operating system:** Linux dunja 5.3.0-40-generic #32-Ubuntu SMP Fri Jan 31 20:24:34 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux

**`git-secret` path:** /usr/bin/git-secret

**`git-secret` version:** 0.3.1

**`git` version:** git version 2.20.1

**Shell type and version:** bash, version 5.0.3(1)-release (x86_64-pc-linux-gnu)

**`gpg` version:** gpg (GnuPG) 2.2.12

joshrabinowitz:
Have `git secret tell`warn about expired, revoked, or otherwise invalid keys. 

Closes #587, #552, #508, #317, #290, #283, and #238


joshrabinowitz:
This PR issues a new warning if it encounters an expired, revoked, or otherwise invalid key:

`git-secret: warning: at least one key for email(s) is revoked, expired, or otherwise invalid: email@example.com`

and also emits a new class of error if the user has multiple keys for an email:

`git-secret: error: N found keys found in gpg keyring for: email@example.com`

joshrabinowitz:
Closing for now, may re-open later.

joshrabinowitz:
Hey @sobolevn, I can't see any reason not to merge this PR, I think this will help a bunch of people with revoked/expired/otherwise bad keys avoid issues with git-secret in the real world.

My only current question about this PR regard the changes at https://github.com/sobolevn/git-secret/pull/553/files#diff-0e013d86c1d0742c9ea18dfddb80e094R542 , which make it so that using `tell` with a user's keychain which has multiple keys under the same email address is a hard fail. Why would you want to `tell` multiple keys from one email address to a git-secret repo? Also, it's not like a user is going to be using `tell` on the same email address repeatedly, unless they're setting up multiple git-secret repos over time.  Using multiple keys for the same email address with git-secret has definitely caused confusion and problems for some users in the real world, so making this an error helps.

On the other hand, it's (technically) non-backward compatible for this unusual corner case, since currently people who were using `tell` on an  email address with multiple _valid_ keys can be blissfully unaware of the fact. So maybe `tell`ing a user with multiple keys should result in a warning, not a failure.

Let me know if you have any thoughts, otherwise I'll merge this one way or the other soon.

sobolevn:
@joshrabinowitz sorry, I have missed this notification!

Looks good to me!

Considering your questions:

> Why would you want to use tell on multiple keys from one email address?

I don't know. The only thing I can think of is having multiple keys for different automation reasons. But, that's not something I have done myself or can recommend doing to other people. 

> so making this an error helps

I agree, but this is a breaking change. And we should point this our in `CHANGELOG.md` and why we have done this. As you said:

> On the other hand, it's (technically) non-backward compatible for this unusual corner case

So, thanks a lot for working on this! 👍

PR #556, opened 2019-10-05
Issue 554 show gpg commands

### LINKED ISSUE #558: stop tests from hanging when using ctrl-c on OSX
As per https://github.com/bats-core/bats-core/issues/155 , there's a known issue with `bats-core` possibly hanging on subprocesses because of it's use of fd3.

We run lots of subprocesses in the tests for  [git-secret](https://github.com/sobolevn/git-secret/), (very few of them what I would consider 'long running', but some that might take a second), but I can't seem to consistently avoid hitting this issue when I use cntrl-C when using bats-core on git-secret on OSX (IE, when interrupting 'make test')-- no matter what I do with fd 3 and trapping SIGINT as suggested in the docs and the bats-core issue above. 

When the tests hang on ctrl-c, they always appear in ps output similarly to this this:

`examp            14622  93.4  0.0  4288452   2132 s002  R    11:44AM   1:29.63 bash /Users/examp/gitsrc/git-secret/vendor/bats-core/libexec/bats-core/bats-exec-test -x /Users/examp/gitsrc/git-secret/tests/test_reveal.bats test_run_-27reveal-27_with_password_argument 1`

and that process appears to be waiting on other subprocesses that have already exited.

I would love to resolve this issue, I've been hitting it for quite a while.

joshrabinowitz:
Originally a Proof Of Concept branch to have git-secret show gpg commands run

Edit: I'm happy with how this came out, I'm cool with merging this, makes verbose modes show many (but not all) of the gpg commands run.

Also, closes #558

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,15 @@
 # Changelog
 
+## {Next Version}
+
+### Features
+
+- Verbose mode shows gpg commands used for hide, reveal, tell, and killperson (#554)
+
+### Misc
+
+- Fix tests from occasional hang when interrupting with cntl-c (#558)
+
 ## Version 0.3.2
 
 ### Bugfixes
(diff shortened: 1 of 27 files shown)


joshrabinowitz:
Note that this was reverted by #559

PR #578, opened 2020-01-23
Allow passing files to be encrypted to git-secret hide

Triodes:
--------------------------------------------------
Allows for passing the names of the files you want to encrypt to `git-secret hide`.
It is possible to reveal specific secrets by passing the relevant file to `reveal`. But when you edit that revealed file and want to encrypt it again you will be presented with errors that the files being encrypted cannot be found. This is because `hide` tries to encrypt all files listed in the fsdb. Even when you have revealed only one file. To circumvent this you can pass the `-F` flag to `hide`. But you will still get a lot of warnings. This pull request aims to solve this problem by allowing you to specify which files you want `hide` to encrypt. `hide` will check if the specified file exists in the database and will abort if it does not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## {Next Version}
+### Features
+- Allow passing paths of files to encrypt to git-secret hide
 
 ### Misc
 
--- a/man/man1/git-secret-hide.1.ronn
+++ b/man/man1/git-secret-hide.1.ronn
@@ -8,7 +8,8 @@ git-secret-hide - encrypts all added files with the inner keyring.
 
 ## DESCRIPTION
 `git-secret-hide` creates an encrypted version (typically called `filename.txt.secret`) 
-of each file added by `git-secret-add` command. 
+of each file added by `git-secret-add` command 
+or the passed `pathspec`s. 
 Now anyone enabled via 'git secret tell' can can decrypt these files. Under the hood,
 `git-secret` uses the keyring in `.gitsecret/keys` and user's secret keys to decrypt the files.
 
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 
 # Folders:
-_SECRETS_DIR=${SECRETS_DIR:-".gitsecret"}   
+_SECRETS_DIR=${SECRETS_DIR:-".gitsecret"}
 # if SECRETS_DIR env var is set, use that instead of .gitsecret
 # for full path to secrets dir, use _get_secrets_dir() from _git_secret_tools.sh
 _SECRETS_DIR_KEYS="${_SECRETS_DIR}/keys"
@@ -12,7 +12,7 @@ _SECRETS_DIR_KEYS_TRUSTDB="${_SECRETS_DIR_KEYS}/trustdb.gpg"
 
 _SECRETS_DIR_PATHS_MAPPING="${_SECRETS_DIR_PATHS}/mapping.cfg"
 
-# _SECRETS_VERBOSE is expected to be empty or '1'. 
+# _SECRETS_VERBOSE is expected to be empty or '1'.
 # Empty means 'off', any other value means 'on'.
 # shellcheck disable=SC2153
 if [[ -n "$SECRETS_VERBOSE" ]] && [[ "$SECRETS_VERBOSE" -ne 0 ]]; then
@@ -42,6 +42,19 @@ BEGIN { FS=":"; OFS=":"; cnt=0; }
   }
 }
 END { if ( cnt > 0 ) print "0"; else print "1"; }
+'
+
+# shellcheck disable=2016
+AWK_FSDB_GET_RECORD='
+BEGIN { FS=":"; OFS=":"; cnt=0; }
+{
+  if ( key == $1 )
+  {
+    cnt++
+    print $0;
+  }
+}
+END { if ( cnt < 1 ) print "1"; }
 '
 
 # shellcheck disable=2016
@@ -125,7 +138,7 @@ function _os_based {
     Linux)
       "$1_linux" "${@:2}"
     ;;
-	
+
     MINGW*)
       "$1_linux" "${@:2}"
     ;;
@@ -247,6 +260,16 @@ function _fsdb_has_record {
 
   # 0 on contains, 1 for error.
   gawk -v key="$key" "$AWK_FSDB_HAS_RECORD" "$fsdb"
+}
+
+function _fsdb_get_record {
+  # First parameter is the key
+  # Second is the fsdb
+  local key="$1"  # required
+  local fsdb="$2" # required
+
+  # returns record if found
+  gawk -v key="$key" "$AWK_FSDB_GET_RECORD" "$fsdb"
 }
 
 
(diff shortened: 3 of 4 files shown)


joshrabinowitz:
@Triodes thank you for this submission.

git-secret specifically discourages using 'hide' to encrypt a subset of files. As the docs for `git secret hide` say:

````
It is recommended to encrypt (or re-encrypt) all the files in a git-secret 
repo each time git secret hide is run.

Otherwise the keychain (the one stored in .gitsecret/keys/*.gpg), may 
have changed since the last time the files were encrypted, and it's 
possible to create a state where the users in the output of git secret 
whoknows may not be able to decrypt the some files in the repo, or 
may be able decrypt files they're not supposed to be able to.

In other words, unless you re-encrypt all the files in a repo each time 
you 'hide' any, it's possible to make it so some files can no longer be
 decrypted by users who should be (and would appear) able to decrypt 
them, and vice-versa.

If you know what you are doing and wish to encrypt or re-encrypt 
only a subset of the files even after reading the above paragraphs, 
you can use the -F or -m option to only encrypt a subset of files. 
The -F option forces git secret hide to skip any hidden files where 
the unencrypted versions aren't present. The -m option skips any 
hidden files that have not be modified since the last time they were encrypted.
````

Now you make an interesting point about how you see lots of warnings when using `-F` - but I also think this is reasonable, taking into account the above paragraphs. Please let us know your thoughts!

Triodes:
Lets see if I understand correctly:
We recommend decrypting and re-encrypting all files to make sure they were encrypted using the latest version of the keychain. This ensures that the hidden files can be decrypted by all those and only those who 'were told'.

Now that makes me curious as to when the keychain actually changes. My assumption was that this only happens when people 'are told' or 'killed'. When that happens all the secrets need to be re-encrypted anyways. 

So under the assumption that we are not 'telling' or 'killing' people, but only modifying the content of the secrets we would not need to re-encrypt every secret just because we edited one right?

Reading this part of the docs also makes me wonder why we would allow partial decrypts using reveal if we recommend always fully decrypting and encrypting.

Lastly to provide a bit of context:
We use git secret to encrypt a git repo which holds the secrets used while configuring our server landscape using the automation tool chef. We have about 700 servers and somewhere between 400 and 500 secret files. When editing one file, decrypting and encrypting just the one is substantially faster and doesn't fill your terminal with 400+ warnings 😄

joshrabinowitz:
@Triodes , thank you for the detailed reply. 

You said:
> So under the assumption that we are not 'telling' or 'killing' people, 
> but only modifying the content of the secrets we would not need to
> re-encrypt every secret just because we edited one right?

That's correct. But git-secret doesn't keep track of what keys were used to encrypt which secrets, so it has no idea if the keys have changed or not since the files were last encrypted.

You also said:
> We have about 700 servers and somewhere between 400 and 500 secret files. 
> When editing one file, decrypting and encrypting just the one is substantially 
> faster and doesn't fill your terminal with 400+ warnings.

Hm. So would it fulfill your desires if there was a way to suppress error messages about _missing unencrypted files_?

Also, @sobolevn , you might be interested in this thread, since it is about using git-secret at decent scale in the real-world.

sobolevn:
Oh wow! This is indeed a decent scale! 

> When editing one file, decrypting and encrypting just the one is substantially faster and doesn't fill your terminal with 400+ warnings

I can understand your point. If I were you I would request the same feature. But.
As @joshrabinowitz have said it is a bad pattern generally. Because people will forget about keychain changes and will a lot of manual work.

@Triodes  What about `-m` option? Does it work for you? I seems to me that it is pretty close to what you are searching: `-m  - encrypt files only when modified.`

joshrabinowitz:
I tested out this case (where there's N secret files, and one is decrypted and changed and you want to re-encrypt the changed one quickly),

`-m` alone won't work for this case because it considers missing files 'changed', and it will abort at the first missing file. (Thus, using `-m -F' is basically the same in this case as using `-F`)

@triodes, @sobolevn 

I think if you want to suppress the N-1 warnings about missing files in your case, that a 'don't show warnings` flags may be what you seek (like I mention above). Interested to hear each of your thoughts.

Triodes:
> -m alone (as suggested in the comment above) won't work for this case because it considers missing files 'changed', and it will abort at the first missing file. (Thus, using -m -F is basically the same in this case as using -F)

I can confirm this

I ran a little test myself and can conclude that revealing all files takes more than a minute. Hiding after that with the `-m` option takes about 30 seconds which is probably due to calculating the hashes.

> I think if you want to suppress the N-1 warnings about missing files in your case, that a 'don't show warnings' flags may be what you seek (like I mention above). Interested to hear each of your thoughts.

I quickly implemented this and tested it. It is about 5 seconds faster, but completely eliminates terminal clutter.

> Reading this part of the docs also makes me wonder why we would allow partial decrypts using reveal if we recommend always fully decrypting and encrypting.

Do you have any thoughts on this?

-----------------------------
### Other thoughts
> So under the assumption that we are not 'telling' or 'killing' people, but only modifying the content of the secrets we would not need to re-encrypt every secret just because we edited one right?

As this assumption is correct: why not force people to (or warn that they should) re-encrypt all the secrets whenever a change to the keychain is made?

joshrabinowitz:
@Triodes asked:
>Reading this part of the docs also makes me wonder why we would allow partial decrypts using reveal if we recommend always fully decrypting and encrypting.

Not forcing people to decrypt all secrets was because of  #253. Also, because of your exact use case - if a repo has many secrets, only one of which is needed for a particular environment, we wanted to avoid forcing the user to expose all secrets. 

@Triodes  also asked:
>why not force people to (or warn that they should) re-encrypt all the secrets whenever a change to the keychain is made?

The -F options to `reveal` and `hide` were also made in response to a specific user request, #253, " Cannot encrypt new files unless all files are decrypted first", in which the user explains "In my case they may not exist as not all contributors will be given access to private keys, but I still want them to be able to encrypt new files."

Also, the docs do say `It is recommended to encrypt (or re-encrypt) all the files in a git-secret repo each time git secret hide is run`, which I assumed would be understood to be necessary if any change to the keychain was made. But you're right, I may be presuming too much about users' understanding of how git-secret works.

If you want to make the docs more verbose about this issue I can go along with that! PR's always welcome, or I'll try to improve the docs soon.

Triodes:
A bit more background:
I do not think the current way git secret works is that disrupting. 30 seconds is not that long.
I thought it would be a nice addition, not knowing the recommendation from the docs.

If this conflicts with the recommendation too much we can also just close this PR.

Due to corporate rules I need to transform this into an RPM (redhat package) each release anyways, so I can always fork the repo, keep it in sync with yours and use my own version to build the RPM.

sobolevn:
@Triodes maybe we can use your PR as a separate plugin?

I mean, your use-case is perfectly valid and I want to preserve it. But, I am also concerned about merging this into the core, because people will use this feature incorrectly and will have troubles with their keychains.

Here's a plugin example: https://github.com/sobolevn/git-secret/wiki/Third-party-plugins

What do you think?

joshrabinowitz:
@Triodes also, have you seen that we build rpms for releases, here? 
https://bintray.com/sobolevn/rpm/git-secret

Triodes:
> @Triodes maybe we can use your PR as a separate plugin?

I have no objection to this. It would be a fine solution for me and others having the same issue.
However, I am probably not proficient enough with shell script to turn this into a plugin as of yet. 

I shall see if I can spare the time to actually learn and do this. A nudge in the right direction would be appreciated 😄

Feel free to use what you want however you see fit.

Triodes:
> @Triodes also, have you seen that we build rpms for releases, here?
https://bintray.com/sobolevn/rpm/git-secret

I was not aware. Thanks!

sobolevn:
@Triodes plugins are easy! Just make executable `git-secret-${yourname}.sh` file and place inside your `$PATH`

joshrabinowitz:
Since we seem to consensus on the situation at present, I'm planning to close this PR soon. 
Feel free to open an issue to discuss the options further.

joshrabinowitz:
Oops closed it prematurely. :]

PR #583, opened 2020-02-07
Update installation.md

inverse:
Add arch installation from AUR


joshrabinowitz:
@sobolevn @inverse I'm not an arch expert, but wouldn't it make more sense to direct people to use pacman?

inverse:
That's why I suggested yay since there is no official AUR helper AFAIK. Yay is bundled with lots of arch based distros as well. Pacman only covers the official arch repos.

inverse:
Aur is not officially supported by arch too but it's used by lots and covers a broader Spectrum.

PR #588, opened 2020-03-06
Fix issue #587

riccardomurri:
Silently skip all GPG identities that have `r` in the second field.

In function `_get_users_in_gpg_keyring`, skip identities that have been revoked.

Issue #587

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -643,13 +643,14 @@ function _get_users_in_gpg_keyring {
   fi
 
   # we use --fixed-list-mode so older versions of gpg emit 'uid:' lines.
-  # here gawk splits on colon as --with-colon, exact matches field 1 as 'uid', and selects field 10 "User-ID" 
-  # the gensub regex extracts email from <> within field 10. (If there's no <>, then field is just an email address 
+  # here gawk splits on colon as --with-colon, exact matches field 1 as 'uid',
+  # skips line if field 2 contains "r" (indicating a revoked identity), and selects field 10 "User-ID"
+  # the gensub regex extracts email from <> within field 10. (If there's no <>, then field is just an email address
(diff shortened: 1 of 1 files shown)


joshrabinowitz:
Hello, @riccardomurri and thank you very much for this PR. !

This is similar to a PR I didn't merge, #553. I'll look at this more and probably merge it soon. 

Meanwhile I welcome you to look at #553 (which I've re-open) and let me know your thoughts.
 
In particular, if we're going to skip 'revoked' keys, we might as well skip 'expired' and otherwise 'invalid' keys. Also it would be nice to let the user know that we're skipping particular keys, and why. 

The reason I didn't merge #553 was that I thought it would cause confusion for people who didn't know (for example) that their key was revoked.

joshrabinowitz:
I merged #533 which issues a warning in this case rather than silently skipping them, so I'm going to close this PR. 

Do you think we should make using expired/revoked/invalid keys a hard error in git-secret?

riccardomurri:
Hello @joshrabinowitz many thanks for your quick reaction on this!

Regarding your question:

> Do you think we should make using expired/revoked/invalid keys a hard error in git-secret?

No, at least "expired" and "revoked" *uid*'s are a normal state of long-lived identities: I have been using GPG since the early 2000's and some of my older email addresses are still part of my public GPG key, but revoked: people can still read and decrypt emails I sent back then, but that uid should not be used for *new* emails.  If `git-secret` errored out hard on revoked identities, then I would need to create a new GPG keypair just for it, and then establish trust, etc. -- which is hard and not necessary.

(Note that I'm talking about revocation of *uid*s in a public key: revoking or expiring the entire key --assuming it's possible, I'm not that expert on GPG-- should be a different business and probably an error, as you suggest. But function `_get_users_in_gpg_keyring` deals with uids not with keys.)

PR #614, opened 2020-09-23
More precise feedback about added files

bbroeksema:
This adapts the output of the add command in order to report exactly which
files have been added. Specially with wildcard patterns, this makes it
easier to verify that expected files are added.

```bash
$ git secret add *.md
git-secret: 0 item(s) added.
$ touch test.md test2.md
$ git secret add *.md
git-secret: Adding test2.md...
git-secret: Adding test.md...
git-secret: 2 item(s) added.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 - Reflect changes in ruby bundler during build process
 - Upgrade build process to ansible 2.9
 - Use shellcheck 0.7.1 with CI, not 'latest' (#609)
+- Improve output of git-secret add
 
 ## Version 0.3.2
 
--- a/man/man1/git-secret-add.1.ronn
+++ b/man/man1/git-secret-add.1.ronn
@@ -3,7 +3,7 @@ git-secret-add - starts to track added files.
 
 ## SYNOPSIS
 
-    git secret add [-i] <pathspec>...
+    git secret add [-v] [-i] <pathspec>...
 
 
 ## DESCRIPTION
@@ -23,6 +23,7 @@ folder using the SECRETS_DIR environment variable.
 
 ## OPTIONS
 
+    -v  - verbose, shows extra information.
     -i  - does nothing, adding paths to .gitignore is now the default behavior.
     -h  - shows this help.
 
--- a/src/commands/git_secret_add.sh
+++ b/src/commands/git_secret_add.sh
@@ -5,11 +5,13 @@ function add {
   local auto_ignore=1
   OPTIND=1
 
-  while getopts "ih" opt; do
+  while getopts "ihv" opt; do
     case "$opt" in
       i) auto_ignore=1;;    # this doesn't change anything
 
       h) _show_manual_for "add";;
+
+      v) _SECRETS_VERBOSE=1;;
 
       *) _invalid_option_for "add";;
     esac
@@ -82,6 +84,8 @@ function add {
 
   local fsdb
   fsdb=$(_get_secrets_dir_paths_mapping)
+  local count
+  count=0
 
   for item in "${items[@]}"; do
     local path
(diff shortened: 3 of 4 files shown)


sobolevn (commented):
I love this idea! Thanks a lot for implementing it.

But, what if there are hundreds of items under `*.md` (as in your example)?
This way we would produce lots of output.

What's your opinion about it?

bbroeksema:
It would match my use case. My assumption is that this is typically a first-case issue, while afterwards the number of new files is smaller. I could live with this one-time information "overload". For all other cases, it would definitely like some explicit feedback about which files have been added.

joshrabinowitz:
Thanks for this feature suggestion and implementation, @bbroeksema !

Here's my input:
* I like showing the count of actually added items using $count, that makes sense
* I think the additional info on each file added should only show when using a new `-v` option, like `git secret hide` has (See <https://github.com/sobolevn/git-secret/blob/master/man/man1/git-secret-hide.1.ronn>)
* I suggest the message should look like "git-secret: adding file: $filename", without ellipses.
   (currently it says it's adding a _key_, and shows ellipses which make it harder to scan the filename)

joshrabinowitz (commented):
Great PR! Please see my comment below.

bbroeksema:
> Thanks for this feature suggestion and implementation, @bbroeksema !

Well, thank you guys for a useful tool and the quick and constructive feedback! Made an effort to improve the MR:

> * I think the additional info on each file added should only show when using a new `-v` option, like `git secret hide` has (See https://github.com/sobolevn/git-secret/blob/master/man/man1/git-secret-hide.1.ronn)

Implemented a `-v` option and added an additional test for with and without the verbose flag.

> * I suggest the message should look like "git-secret: adding file: $filename", without ellipses.
>   (currently it says it's adding a _key_, and shows ellipses which make it harder to scan the filename)

Fair enough. Removed the ellipses. In fact I changed the current MR to ` _message "adding file: $key"`, where, as far as I understand, $key *is* the filename (the output also suggests this). Not taking in account the `-v` yet though,

Additionally:
* Updated `CHANGELOG.md`
* Updated `man/man1/git-secret-add.1.ronn -> I'm not sure about `man/man1/git-secret-add.1`. That looks generated, I didn't figure out, how though.

joshrabinowitz:
hello @bbroeksema , 

1) you are correct about $key, I misread. 

2) Tests shouldn't look for exact output line-by-line; setting SECRETS_TEST_VERBOSE will cause these tests to fail. See other tests where we look for particular output lines while ignoring SECRETS_TEST_VERBOSE output for how we do this.

3) Yes, the non-.ronn versions of the man pages are generated by `make build-man`

I'll look at this in more depth and test it some, and may make a few more suggestions, or just merge your PR and apply changes later. Thank you again!

bbroeksema:
Hi @joshrabinowitz, re `2.`: what do you suggest should be tested there? I could do something similar as in the `hide` tests: `[[ "$output" == *"git-secret: done. 1 of 1 files are hidden."* ]]`. I.e. using wildcards and testing the "main" message. But testing the main message is also done in the non-verbose test. I'd appreciate some more concrete input to move this forward.

joshrabinowitz:
@bbroeksema yes, that's what I recommend.

The idea is that in verbose, (and non-verbose modes), we test that the expected _non-verbose_ text is emitted.  

You can also test for the expected addedverbose-mode text if `git-secret` is run in verbose mode. For example, you could add something like this for verbose mode to test that you're getting additional lines as expected.

    [[ "$output" == *"git-secret: adding file: ${TEST_DEFAULT_FILENAME}"* ]]

bbroeksema:
@joshrabinowitz 6 days :) life is a b*tch sometimes. Changed the test as suggested.

PR #626, opened 2020-11-21
add bash string escaping to _add_ignored_file

friedenberg:
running `git secret add [FILENAME]` where `[FILENAME]` contains special characters adds a filename to `.gitignore` that is not properly escaped

this change uses `printf '%q'` to correctly escape the filename before adding it to `.gitignore`

No

No

Any other comments?
-------------------
No


sobolevn (approved):

sobolevn:
Thanks!

Issue #631, opened 2020-12-28
store '-armor' versions of .secret files instead of binary versions, so they're ascii and not binary


joshrabinowitz:
In #630, @jpbochi suggested he'd "personally prefer to commit armored files rather than binaries", which is a good idea.

joshrabinowitz:
I think this may fix #136

jonjensen:
I mentioned this in a pending PR already - https://github.com/sobolevn/git-secret/pull/661#discussion_r627719375 - but I prefer the current behavior of storing as binary by default. Seems like --armor would be a nice user-configurable option, but binary seems the better default. Has anyone mentioned actual problems with storing binary blobs in Git as git-secret has always done before?

sobolevn:
Done!

PR #632, opened 2020-12-28
use gpg --armor when hiding/encrypting files

joshrabinowitz:
For #631: 

Use gpg --armor when encrypting/hiding files, so secret files are text instead of binary.

When testing with gpg 2.0.22 and git-secret master,  I found that I could 'reveal' files encrypted whether or not they had been hidden with the --armor option.

This needs to be tested with older and newer versions of gpg


jpbochi:
Hopefully, this won't be a breaking change, since `gpg` is able to decrypt files regardless of them being in binary or armored text formats.

I do like the idea of making armored text be the default format. The only advantage of binary is that the files would be smaller, but that's not a big difference for files that are typically small.

joshrabinowitz:
I think this may fix #136

joshrabinowitz:
@sobolevn Also I rebased master into this branch, but that might not be the right thing (itnow shows over 300 files changed). let me know if I should recreate this branch or something else

sobolevn:
@joshrabinowitz yes, I think that you should create a new branch and cherry-pick the required commits from `joshrabinowitz:issue-630-hide-armored` into it! 👍

joshrabinowitz:
See replacement PR #661

joshrabinowitz:
This PR is replaced by #661

PR #637, opened 2021-01-13
Issue 634 tell duplicate emails

### LINKED ISSUE #634: change 'tell' to give an error if you tell the same email address twice, so it works more clearly
Issue #633 documents an issue where a user (apparently) added two different keys, with the same email address, to their git-secret repo. 

In my testing with 0.3.3, I could not add the same key twice to a git-secret repo (when I used `git secret tell email@example.com` the second time, it appeared to add the key again, but only one key for the email appears in the gitsecret keyring in `.gitsecret/keys`)

If you enable verbosity with `SECRETS_VERBOSE=1 `, you get results that show the key is not re-added; However, it does _appear_ to add the key for the user a second time. 
(If you don't use the verbose mode, you don't see anything about the existing key.)

````
% SECRETS_VERBOSE=1 ../git-secret/git-secret tell bobby-test@example.com

gpg: key 8030A3F3: "bobby-test (bobby-test comment) <bobby-test@example.com>" not changed
gpg: Total number processed: 1
gpg:              unchanged: 1
git-secret: done. bobby-test@example.com added as user(s) who know the secret.
````
 Instead, git-secret should report an error like 

`git-secret: abort: there is already a key for the email blah@example.com`

joshrabinowitz:
Closes #634 by returning an error if you try to `tell` an  email address that currently has access to the git-secret repo.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,10 +5,12 @@
 ### Bugfixes
 
 - Escape filenames with special characters before adding to .gitignore
+- Better error handling around telling an email twice (#634)
 
 ### Misc
 
 - Improve documentation about releases
+- Man page improvements
 
 ## Version 0.3.3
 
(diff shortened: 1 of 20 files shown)


joshrabinowitz:
@sobolevn can you please review this PR?

Issue #640, opened 2021-01-19
extra new lines in paths/mapping.cfg read as candidate files to git secret reveal


otosky:
So I know that ideally extra newlines wouldn't exist in the mapping.cfg file, but encountered a situation in my org where this slipped through the cracks.  

Error message didn't immediately make it clear why a file named `<ROOT_PATH>.secret` was trying to be processed by `reveal` command.

What are the steps to reproduce this issue?
-------------------------------------------
1. Add extra new lines to end of `.gitsecret/paths/mapping.cfg`
2. Run `git secret reveal -f`

What happens?
-------------
gpg tries to operate on a file `<ROOT_PATH>.secret` that doesn't exist

What were you expecting to happen?
----------------------------------
Errant new lines in `mapping.cfg` to be skipped or a warning emitted stating that filename presented to `reveal` is an empty string.

Any logs, error output, etc?
----------------------------

Any other comments?
-------------------

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`) 
Linux 51dfe65a70cf 4.19.121-linuxkit #1 SMP Tue Dec 1 17:50:32 UTC 2020 x86_64 GNU/Linux

**`git-secret` path:** (`which git-secret`)
/usr/bin/git-secret

**`git-secret` version:** (`git secret --version`)
0.3.2

**`git` version:** (`git --version`)
git version 2.11.0

**Shell type and version:** (`$SHELL --version`)
GNU bash, version 4.4.12(1)-release (x86_64-pc-linux-gnu)

**`gpg` version:** (`gpg --version`)
gpg (GnuPG) 2.1.18
libgcrypt 1.7.6-beta

joshrabinowitz:
Is there a way git-secret itself will create extra blank lines in mapping.cfg?

Or is this just protecting against user errors when writing mapping.cfg manually (which is highly discouraged)?

otosky:
I'll have to double check what produced the extra newline artifact in the first place.  I only noticed the issue after a coworker committed some code to a shared repo that had added some entries to the mapping.cfg file.

I'll poke around to see whether I can reproduce the error from `git-secret` itself.  I don't think he would have had any reason to explicitly alter the cfg file, but it could very well have been a manual intervention after running the proper git secret commands.

PR #641, opened 2021-01-19
Add check to skip errant new lines in path mapping

otosky:
Add a check to array of mapping.cfg records to filter out empty-strings/newlines.

Current implementation doesn't emit any warning that errant new lines in `mapping.cfg` will be processed in `git secret reveal` as `<ROOT_DIR>.secret`.

Ideally the mapping file wouldn't contain any extra new lines, but wanted to address the edge case where it does.

https://github.com/sobolevn/git-secret/issues/640

Any other comments?
-------------------
Deferring to maintainer opinion on whether a debug message would be helpful here so that user's are warned that their `mapping.cfg` file contains unnecessary newlines.

### DIFF
--- a/src/commands/git_secret_reveal.sh
+++ b/src/commands/git_secret_reveal.sh
@@ -47,6 +47,11 @@ function reveal {
 
   if [ ${#to_show[@]} -eq 0 ]; then
     while read -r record; do
+
+      if [[ "$record" == "" ]]; then
+        continue # skip new lines / empty strings
+      fi
+
       to_show+=("$record")  # add record to array
     done < "$path_mappings"
   fi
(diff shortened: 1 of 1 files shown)


sobolevn:
Hi, @otosky!

Sorry for the long reply, but could please clarify how empty lines can end up in `mapping.cfg`?
I would love to merge this if we can create a reproducable test case.

otosky:
Hey @sobolevn, also apologies for the late reply!

The empty lines I was encountering were a product of a teammate manually editing the `mapping.cfg` file, which I understand is not best practice.  So this merge might actually be moot.

sobolevn:
Ok, thanks a lot for the clarification! Let's close this for now. But, if you have this problem once again, I would be happy to merge a similar PR with some tests to back it up.

👍

PR #654, opened 2021-04-14
Add security disclaimer for git-secret-killperson specifying what is and is not readable by a user after having been removed from the repository's keyring

### LINKED ISSUE #653: Request for clarification: Access to historical secrets after revocation
I wasn't able to find an answer to this question on the website or the documentation, and was not able to decipher definitively one way or the other from the source code, so I pose my question here:

What, if any, measures does git-secret take to prevent post-revocation access to historical secrets in the git history?

Assuming a user has access to a public repository, has been included in the keyring and that the secrets in the repository have been re-hidden using this keyring. If the user is then "killed", what prevents them from checking out an older reference and accessing the secrets as they were at that point?

I agree that the example may be a little contrived, given the decentralized nature of git which means that:
- Even if steps are taken within the remote to rebase or filter out the older keyrings and encrypted secrets, a copy may still exist on their local machine
- That in most cases a repository containing secrets like this will likely be private in some form, meaning the revocation in most cases would be coupled with restricting access to the repository itself.

But I did find it somewhat concerning that this was not addressed at all, as far as I could tell. [git-crypt](https://github.com/AGWA/git-crypt#limitations) for example has a very clear disclaimer for this type of scenario, and the lack of it might provide a false sense of security for users of git-secret, which might skew decision-making when choosing an in-tree encryption method.

MathiasPius:
Adds security disclaimer for `git-secret-killperson` specifying what is and is not readable by a user after having been removed from the repository's keyring.

Closes #653

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ### Misc
 
+- Add security disclaimer for git-secret-killperson
 - Improve documentation about releases
 - Man page improvements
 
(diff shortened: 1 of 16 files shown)


sobolevn (approved):
Thanks!

PR #661, opened 2021-05-04
Issue 630 hide armored2

joshrabinowitz:
Recreation of #632 
For #631 

Use gpg --armor when encrypting/hiding files, so secret files are text instead of binary.

When testing with gpg 2.0.22 and git-secret master, I found that I could 'reveal' files encrypted whether or not they had been hidden with the --armor option.

This needs to be tested with older and newer versions of gpg, also to make sure armored files encrypted with 2.2 are decryptable with ancient gnupg versions.

### DIFF
--- a/man/man1/git-secret-hide.1.md
+++ b/man/man1/git-secret-hide.1.md
@@ -12,7 +12,7 @@ of each file added by `git-secret-add` command.
 Now anyone enabled via 'git secret tell' can can decrypt these files. Under the hood,
 `git-secret` uses the keyring in `.gitsecret/keys` and user's secret keys to decrypt the files.
 
-It is recommended to encrypt (or re-encrypt) all the files in a git-secret repo each
+It is recommended to encrypt (or re-encrypt) all the files in a `git-secret` repo each
 time `git secret hide` is run.
 
 Otherwise the keychain (the one stored in `.gitsecret/keys/*.gpg`),
@@ -21,15 +21,17 @@ to create a state where the users in the output of `git secret whoknows`
 may not be able to decrypt the some files in the repo, or may be able decrypt files
 they're not supposed to be able to.
 
-In other words, unless you re-encrypt all the files in a repo each time you 'hide' any,
+In other words, unless you re-encrypt all the files in a repo each time you `hide` any,
 it's possible to make it so some files can no longer be decrypted by users who should be
 (and would appear) able to decrypt them, and vice-versa.
 
-If you know what you are doing and wish to encrypt or re-encrypt only a subset of the files
-even after reading the above paragraphs, you can use the `-F` or `-m` option to only encrypted
-a subset of files. The `-F` option forces `git secret hide` to skip any hidden files
-where the unencrypted versions aren't present. The `-m` option skips any hidden files that have
-not be modified since the last time they were encrypted.
+If you know what you are doing and wish
+to encrypt or re-encrypt only a subset of the files
+even after reading the above paragraphs, you can use the `-F` or `-m` options.
+The `-F` option forces `git secret hide` to skip any hidden files
+where the unencrypted versions aren't present.
+The `-m` option skips any hidden files that have
+not be been modified since the last time they were encrypted.
 
 
 ## OPTIONS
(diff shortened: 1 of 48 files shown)


src/commands/git_secret_hide.sh
```diff
@@ -162,7 +162,7 @@ function hide {
       # encrypt file only if required
       if [[ "$update_only_modified" -eq 0 ]] || [[ "$fsdb_file_hash" != "$file_hash" ]]; then
 
-        local args=( --homedir "$secrets_dir_keys" "--no-permission-warning" --use-agent --yes "--trust-model=always" --encrypt )
+        local args=( --homedir "$secrets_dir_keys" "--no-permission-warning" --use-agent --yes "--trust-model=always" --encrypt --armor )
```
sobolevn, line 165: What potential side effects on existing users / files it might have? 🤔

sobolevn (approved):
Thanks! Small and very useful change! 👍

sobolevn:
I think that it would be better to merge this after I finish with `0.4.0` release. Because `rpm` and `deb` packages are already published. I don't want alpine package to be different.

I will merge it later on! Thanks again!

src/commands/git_secret_hide.sh
joshrabinowitz: as long as the armored files are fully interoperable between versions of gnupg, then the differences are
* the secret files are text, not binary
* as result they are larger in file size
* maybe git will show diffs in the armored txt files if you do something like 'git log -p' on them, instead of just saying something like 'binary files differ'

Note also that this feature was requested by a user.

sobolevn: I think that this might be a good candidate for configuration 🤔

jonjensen: @sobolevn Agree on this being a configuration option. For me, binary by default is more sensible. There's no reason I would want these ASCII armored. Benefits of binary: small, diff doesn't try to show changes, and why not? Git supports binary well.

sobolevn:
Ok, I am going to add a configuration option and merge this into `0.5.0`

sobolevn:
@joshrabinowitz done! I would appreciate your feedback! 👍

sobolevn (approved):
LGTM, I will wait for @joshrabinowitz

man/man1/git-secret-hide.1.md
```diff
@@ -7,37 +7,29 @@ git-secret-hide - encrypts all added files with the inner keyring.
...
-Also, it is possible to modify the names of the encrypted files by setting `SECRETS_EXTENSION` variable.
-
-(See [git-secret(7)](http://git-secret.io/git-secret) for information about renaming the .gitsecret
-folder using the SECRETS_DIR environment variable.
-
-You can also enable verbosity using the SECRETS_VERBOSE environment variable,
-as documented at [git-secret(7)](http://git-secret.io/)
+If you know what you are doing and wish to encrypt or re-encrypt only a subset of the files
+even after reading the above paragraphs, you can use the `-F` or `-m` option to only encrypted
+a subset of files. The `-F` option forces `git secret hide` to skip any hidden files
```
joshrabinowitz, line 30: we should remove "to only encrypted (sic) a subset of files." from the end of the sentence starting with "If you know what you are doing", and replace it with "to do so", because it repeats what we say in start of sentence.

man/man1/git-secret-hide.1.md
```diff
@@ -7,37 +7,29 @@ git-secret-hide - encrypts all added files with the inner keyring.
...
+where the unencrypted versions aren't present. The `-m` option skips any hidden files that have
+not be modified since the last time they were encrypted.
```
joshrabinowitz, line 32: change to "not _been_ modified"

man/man1/git-secret-hide.1.md
```diff
@@ -50,13 +42,24 @@ as documented at [git-secret(7)](http://git-secret.io/)
...
+
+## ENV VARIABLES
+
+- `SECRETS_GPG_COMMAND` to change the default `gpg` command to anything else
+- `SECRETS_GPG_ARMOR` boolean to enable [`--armor` mode](https://www.gnupg.org/gph/en/manual/r1290.html) to store secrets in text format over binary
+- `SECRETS_DIR` to change the default `.gitsecret/` folder to another name as documented at [git-secret(7)](http://git-secret.io/)
+- `SECRETS_EXTENSION` to change the default `.secret` file extension
+- `SECRETS_VERBOSE` to change the output verbosity as documented at [git-secret(7)](http://git-secret.io/)
+- `SECRETS_PINENTRY` to change the [`gpg --pinentry` mode](https://github.com/gpg/pinentry) as documented at [git-secret(7)](http://git-secret.io/)
+
```
joshrabinowitz, line 54: maybe alter the text from "to change" to "changes"? Reads a little better.

src/_utils/_git_secret_tools.sh
```diff
@@ -29,7 +29,19 @@ fi
 : "${SECRETS_OCTAL_PERMS_COMMAND:="_os_based __get_octal_perms"}"
 : "${SECRETS_EPOCH_TO_DATE:="_os_based __epoch_to_date"}"
 
-# Temp Dir
+# _SECRETS_GPG_ARMOR is expected to be empty or '1'.
+# Empty means 'off', any other value means 'on'.
+# See: https://github.com/sobolevn/git-secret/pull/661
+# shellcheck disable=SC2153
+if [[ -n "$SECRETS_GPG_ARMOR" ]] && [[ "$SECRETS_GPG_ARMOR" -ne 0 ]]; then
```
joshrabinowitz, line 36: I'd prefer not to use a separate `_SECRETS_GPG_ARMOR` env var, and instead just set the correct param into $args at line 169 below using `SECRETS_GPG_ARMOR`
Seems like simpler logic and less variables

sobolevn: 👍

sobolevn:
Thanks a lot!

PR #685, opened 2021-06-15
Rename `killperson` to `removeperson`

### LINKED ISSUE #684: Reconsider "killperson" command name
As a person who just had to offboard a colleague who is terminally ill, having to enter a `killperson` command to remove their access to encrypted secrets leaves a very bad taste in my mouth.

Would you consider changing the command to be `removeperson`, `removeaccess` or similar which hopefully should be less offensive, and possibly keep `killperson` as an invisible alias?

Tenzer:
This renames the `killperson` command to `removeperson`. `killperson` is unnecessarily hostile, as noted in #684.

Yes, this fixes #684.

None

Any other comments?
-------------------
I noticed the contribution guidelines talked about installing pre-commit hooks which no longer seem to be in the repository, so I removed the references to that as well.


src/commands/git_secret_removeperson.sh
```diff
@@ -1,14 +1,14 @@
 #!/usr/bin/env bash
 
 
-function killperson {
+function removeperson {
```
sobolevn, line 4: Let's add something like:

```bash
function killperson {
   echo 'WARNING: `killperson` is renamed to `removeperson`, it will be removed in the next version, please upgdare!'

   remove_person $@
}
```

tests/test_removeperson.bats
```diff
@@ -57,15 +57,15 @@ function teardown {
 }
 
 
-@test "run 'killperson' with multiple arguments" {
+@test "run 'removeperson' with multiple arguments" {
```
sobolevn, line 60: Please, add a single test that `killperson` is still working, we don't want to break it just yet.

sobolevn (changes requested):
Thanks a lot for your work! Really impressive!

src/commands/git_secret_removeperson.sh
Tenzer: Sure, I have added that in b186d6c8deed4290b77b1117e61c80cbcbf44e84.

tests/test_removeperson.bats
Tenzer: Added in b186d6c8deed4290b77b1117e61c80cbcbf44e84.

src/commands/git_secret_removeperson.sh
joshrabinowitz: I think we should leave` killperson` in indefinitely, and not deprecate it, so old scripts continue to work as expected.
I'm fine with updating all the docs to not mention `killperson` any more though.

Also while we're considering this, it's always bugged me a little that you `tell` a user and not `tellperson` or `addperson` (which would be more symmetric with `removeperson`). 

What do you all think about renaming the command  `addperson` (still supporting `tell` of course) and stop documenting `tell`?

sobolevn: @joshrabinowitz I have changed the warning message. I think that we are free not to do this at the next version, but just sometime in the future.

Let's open a new issue for your `tellperson` proposal.

sobolevn (approved):
Thanks! Great work!

Issue #710, opened 2021-08-19
git secret cat not working when cd'ed to subdir


SadriG91:
What are the steps to reproduce this issue?
-------------------------------------------
1. git secret init
2. git secret tell my@email.com
3. git secret add file
4. git secret hide
5. git secret cat file 

(Note added by @joshrabinowitz : the steps above are not sufficient to replicate this bug. See my comment below for steps to replicate.)

What happens?
-------------
decrypt_message failed: No such file or directory 

What were you expecting to happen?
----------------------------------
cat encrypted file to stdout

Any logs, error output, etc?
----------------------------
(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)
gpg: can't open '/path/to/repo/file.secret': No such file or directory
gpg: decrypt_message failed: No such file or directory
git-secret: abort: problem decrypting file with gpg: exit code 2: /path/to/repo/file.secret

Any other comments?
-------------------
Even thought i am in the correct folder where secret file is once i run the command it shows the root_of_repo/file.secret instead root_of_repo/path/to/file.secret
gpg: can't open '/path/to/repo/file.secret': No such file or directory
but the file is in '/path/to/repo/folder/keys/file.secret'

What versions of software are you using?
----------------------------------------
**Operating system:** (`uname -a`) Darwin C02F70UQML7L 20.6.0 Darwin Kernel Version 20.6.0: Wed Jun 23 00:26:31 PDT 2021; root:xnu-7195.141.2~5/RELEASE_X86_64 x86_64

**`git-secret` path:** (`which git-secret`) /usr/local/bin/git-secret

**`git-secret` version:** (`git secret --version`) 0.4.0

**`git` version:** (`git --version`) git version 2.31.0

**Shell type and version:** (`$SHELL --version`) zsh 5.8 (x86_64-apple-darwin20.0)

**`gpg` version:** (`gpg --version`) gpg (GnuPG) 2.3.1

sobolevn:
Are you sure that you don't have any other files named `file` in the root of your repo? Seems like a path / name conflict to me.

You can always submit a failing test: https://github.com/sobolevn/git-secret/blob/master/tests/test_cat.bats
It would be a great help! 👍

joshrabinowitz:
I think I understand what this is about. There's a bug with `git secret cat`

The hint is where OP talks about `git secret cat` using  `repo/file.secret instead repo/path/to/file.secret`. 

To replicate:

````
mkdir repo
cd repo
[...]

PR #718, opened 2021-08-30
start test for git-secret cat bug #710

joshrabinowitz:
Start of test showing bug #710, and fix for same


joshrabinowitz:
Closing, may reopen later.

Issue #722, opened 2021-10-15
Makefile install error 127 when running MINGW64


bennycode:
What are the steps to reproduce this issue?
-------------------------------------------
1. Open [Git Bash](https://gitforwindows.org/) (MINGW64) on Windows 10 Pro
2. Run `git clone https://github.com/sobolevn/git-secret.git git-secret`
3. Run `cd git-secret`
4. Run `make build`
5. Run `PREFIX="/usr/local" make install`

What happens?
-------------

Error 127 shows up:

```bash
bn@BENNY-RYZEN MINGW64 /c/dev/projects/git-secret (master)
$ make build
make: Nothing to be done for 'build'.

bn@BENNY-RYZEN MINGW64 /c/dev/projects/git-secret (master)
$ PREFIX="/usr/local" make install
C:/Program Files/Git/usr/bin/sh.exe ./utils/install.sh "C:/Program Files/Git/usr/local"
/usr/bin/sh: C:/Program: No such file or directory
make: *** [Makefile:25: install] Error 127
```

What were you expecting to happen?
----------------------------------

The `git-secret` tool should be installed successfully.

What versions of software are you using?
----------------------------------------
**Operating system:** MINGW64_NT-10.0-19042 BENNY-RYZEN 3.1.7-340.x86_64 2020-10-23 13:08 UTC x86_64 Msys (Windows 10 Pro)

**`git-secret` path:** no git-secret

**`git-secret` version:** no git-secret

**`git` version:** git version 2.29.2.windows.3

**Shell type and version:** GNU bash, version 4.4.23(1)-release (x86_64-pc-msys)

**`gpg` version:** gpg (GnuPG) 2.2.25

sobolevn:
Ok, then the problem is in `build`, not in `git-secret` 🤔 

For now, you can just replace `make build` with `make git-secret`. But, I will provide a fix right now! 👍

bennycode:
Hi @sobolevn and thank you for taking immediate action. Unfortunately, I still cannot install `git-secret`.

Here is what I did:

**1. Running `git pull`:**

```
$ git pull
remote: Enumerating objects: 3, done.
remote: Counting objects: 100% (3/3), done.
remote: Compressing objects: 100% (3/3), done.
remote: Total 3 (delta 0), reused 0 (delta 0), pack-reused 0
Unpacking objects: 100% (3/3), 2.60 KiB | 532.00 KiB/s, done.
From https://github.com/sobolevn/git-secret
[...]

PR #724, opened 2021-10-15
Support spaces in shell path

bennycode:

On Windows with MINGW64 the `${SHELL}` command can become `C:/Program Files/Git/usr/local` so it must be put in quotation marks to support the space character in "Program Files".

Yes: https://github.com/sobolevn/git-secret/issues/722

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,4 @@
+SHELL:=bash
 PREFIX?="/usr"
 DESTDIR?=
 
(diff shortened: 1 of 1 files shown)


sobolevn:
Maybe something like this? https://stackoverflow.com/questions/4058840/makefile-that-distinguishes-between-windows-and-unix-like-systems

sobolevn:
Looks like it works! @bennycode can you please try your local setup?
Thanks a lot for working on this!

bennycode:
Just using `bash ./utils/tests.sh` works on Windows but what was the reason for it to be `/usr/bin/env bash`?

PR #766, opened 2022-02-03
Bdh cleanup

bryan-howard:
Removes trailing whitespace.
Addresses all outstanding shellcheck warnings.

I don't think so.

No.

Any other comments?
-------------------
No.

### DIFF
--- a/src/main.sh
+++ b/src/main.sh
@@ -26,13 +26,6 @@ function _check_setup {
 }
 
 
-function _incorrect_usage {
-  echo "git-secret: abort: $1"
-  usage
-  exit "$2"
-}
-
-
 function _show_version {
   echo "$GITSECRET_VERSION"
   exit 0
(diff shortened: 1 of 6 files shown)


src/main.sh
```diff
@@ -41,7 +34,9 @@ function _show_version {
 
 function _init_script {
   if [[ $# == 0 ]]; then
-    _incorrect_usage 'no input parameters provided.' 126
```
sobolevn, line 44: Why did you change this?

joshrabinowitz:
@sobolevn @bryan-howard I'm planning to merge this soon, then revert the `_incorrect_usage` call commented on above.

src/main.sh
bryan-howard: shellcheck correctly warns that the usage() function expects parameters but is never called with any. I have just submitted a different pull request based on the current master which passes the rest of the parameter list to the usage function rather than inline it. So you can forget this part of this submission.

PR #779, opened 2022-02-20
show and fix 'style' complaints from shellcheck

joshrabinowitz:
show and fix shellcheck 'style' complaints


sobolevn (approved):
Awesome! Love it! 👍 
Thanks for fixing this.

PR #791, opened 2022-03-15
fix writing .gitignore when using multiple files with `add`

joshrabinowitz:
WIP #789

### DIFF
--- a/tests/test_add.bats
+++ b/tests/test_add.bats
@@ -23,7 +23,7 @@ function teardown {
   # Preparations:
   local filename="$TEST_DEFAULT_FILENAME"
   echo "content" > "$filename"
-  echo "$filename" > ".gitignore"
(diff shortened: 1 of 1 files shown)


src/commands/git_secret_add.sh
```diff
@@ -74,6 +74,9 @@ function add {
       # see https://github.com/sobolevn/git-secret/issues/18 for more.
       _message "$message"
       _message "auto adding them to .gitignore"
+
+      # we need changes below for #789
+      # because this adds multiple filenames on one line
       for item in "${not_ignored[@]}"; do
```
sobolevn, line 80: I guess that array expansion here does not work correctly for some reason 🤔 

Because `_add_ignored_file` looks correct to me: https://github.com/sobolevn/git-secret/blob/92d47c218de4e80da9f5a2681d23fe019a645520/src/_utils/_git_secret_tools.sh#L329

Issue #806, opened 2022-03-28
"gpg: can't connect to the agent: IPC connect call failed" error in docker alpine on shared volume


paslandau:
I try to build an alpine docker image that contains `git-secret` and can be used to encry/decrypt files in a codebase that is shared with the host system.

When running `git secret killperson` in the container within the shared folder, the error

```
gpg: can't connect to the agent: IPC connect call failed
```

is shown. It does not appear when `git secret` is used in a non-shared folder.

![git-secret-error](https://user-images.githubusercontent.com/7747310/160389468-34430a46-c8b6-49da-bd3a-6881ba815f28.gif)

What were you expecting to happen?
----------------------------------
No error is shown

What are the steps to reproduce this issue?
-------------------------------------------
1. Build a docker image to reproduce the issue

```
docker build -t my-git-secret -<<'EOS'

FROM alpine:3.15.2 as base

# install git-secret
# @see https://git-secret.io/installation#alpine
ADD https://gitsecret.jfrog.io/artifactory/api/security/keypair/public/repositories/git-secret-apk /etc/apk/keys/git-secret-apk.rsa.pub

RUN echo "https://gitsecret.jfrog.io/artifactory/git-secret-apk/all/main" >> /etc/apk/repositories  && \
    apk add --update --no-cache \
        git \
        git-secret \
        gnupg

# create GPG keys     
RUN touch /tmp/batch.txt && \
    echo "Key-Type: 1" | tee -a "/tmp/batch.txt" && \
    echo "Key-Length: 2048" | tee -a "/tmp/batch.txt" && \
    echo "Subkey-Type: 1" | tee -a "/tmp/batch.txt" && \
    echo "Subkey-Length: 2048" | tee -a "/tmp/batch.txt" && \
    echo "Name-Real: John Doe" | tee -a "/tmp/batch.txt" && \
    echo "Name-Email: john.doe@example.com" | tee -a "/tmp/batch.txt" && \
    echo "Expire-Date: 0" | tee -a "/tmp/batch.txt" && \
    echo "%no-protection" | tee -a "/tmp/batch.txt" && \
    cat /tmp/batch.txt | gpg --batch --gen-key 
    
EOS
```

- uses `alpine:3.15.2` as base image
- installs `git, gnupg and git-secret`
- creates a gpg key pair for `john.doe@example.com`

2. create an empty folder on the host system
```
mkdir shared
```

3. Start a container with the image, share the folder and attach a shell session
```
[...]

PR #812, opened 2022-03-31
in verbose mode, show some gnupg perm warnings

### LINKED ISSUE #811: reconsider our use of `--no-permission-warning` because sometimes it is hiding actual permission problems
many times when git-secret is interacting with gnupg, we use the `--no-permission-warning` option. 

We periodically see issues which are related to permissions and I suspect that some of these issues could be spotted earlier by git-secret not using `--no-permission-warning`

This also relates to possibly setting permissions more carefully and/or the user's `umask`

joshrabinowitz:
Closes #811 

may relate to #806 

@sobolevn : Maybe we should just not use `--no-permission-warnings` in these cases at all, (even when in non-verbose mode).

### DIFF
--- a/tests/_test_base.bash
+++ b/tests/_test_base.bash
@@ -254,7 +254,7 @@ function set_state_initial {
 function set_state_git {
   local has_initial_branch_option
   has_initial_branch_option=$(is_git_version_ge_2_28_0) # 0 for true
-  if [ "$(has_initial_branch_option)" eq 0 ]; then
+  if [[ "$has_initial_branch_option" == 0 ]]; then
     git init --initial-branch=main >> "$TEST_OUTPUT_FILE" 2>&1
   else
     git init >> "$TEST_OUTPUT_FILE" 2>&1
--- a/tests/_test_base.bash
+++ b/tests/_test_base.bash
@@ -252,7 +252,7 @@ function set_state_initial {
 
 
 function set_state_git {
-  if [ "$(is_git_version_ge_2_28_0)" eq 1 ]; then
+  if [ "$(is_git_version_ge_2_28_0)" eq 0 ]; then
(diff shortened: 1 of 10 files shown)


tests/_test_base.bash
```diff
@@ -239,7 +256,13 @@ function set_state_initial {
 
 
 function set_state_git {
-  git init >> "$TEST_OUTPUT_FILE" 2>&1
+  local has_initial_branch_option
+  has_initial_branch_option=$(is_git_version_ge_2_28_0) # 0 for true
+  if [[ "$has_initial_branch_option" == 0 ]]; then
+    git init --initial-branch=main >> "$TEST_OUTPUT_FILE" 2>&1
```
sobolevn, line 262: Why do we set `main` as the default branch?

sobolevn (commented):
Yes, I think that disabling permissions warning can indeed backfire.
This is a good start!

tests/_test_base.bash
joshrabinowitz: This is to avoid these warnings from git and to avoid the default branch name 'master': 

````
hint: Using 'master' as the name for the initial branch. This default branch name
hint: is subject to change. To configure the initial branch name to use in all
hint: of your new repositories, which will suppress this warning, call:
hint: git config --global init.defaultBranch <name>
hint: 
hint: Names commonly chosen instead of 'master' are 'main', 'trunk' and
hint: 'development'. The just-created branch can be renamed via this command:
hint: 
git branch -m <name>
````

joshrabinowitz: See also https://www.theserverside.com/feature/Why-GitHub-renamed-its-master-branch-to-main

sobolevn: Ok 👍

joshrabinowitz:
@sobolevn let me know if you're ok with me merging this.

PR #837, opened 2022-04-21
describe Windows installation

bennycode:
Adding details for Windows users to: https://git-secret.io/installation


sobolevn (commented):
This is fine by me, because I have no idea about how windows works.

bennycode:
I just reinstalled Windows 11 and `git-secret` on my PC, so I had to go through these steps and thought it would be nice to add them to the official docs. :)

sobolevn (approved):

sobolevn:
I will let @joshrabinowitz to merge this, maybe provide some extra feedback 🙂

joshrabinowitz:
Can we add mention alternatives like wsl, cygwin, and mingw-w64? If the docs could explain the differences a little bit too, and why someone would choose one or another path, that would be great. It seems like there's a bunch of ways to go to use git-secret on windows.

PR #880, opened 2022-06-05
Update Dockerfile

sobolevn:


sobolevn:
@joshrabinowitz looks like JFrog works now!

Issue #881, opened 2022-06-06
release 0.5.1 for all platforms except alpine, so we can pass self-tests with the release


joshrabinowitz:
Idea:  remove support for alpine packages from git-secret and release 0.5.1; then when we resolve the alpine issue(s) we release a new 0.5.2 
@sobolevn

joshrabinowitz:
@sobolevn I've removed apk building from the release process for now, can you release 0.5.1 which should pass all self-tests now

PR #882, opened 2022-06-10
temp removal of alpine releases for #881

joshrabinowitz:
temporary removal of alpine releases for #881


joshrabinowitz:
recreating based on current master

PR #915, opened 2022-07-28
avoid pacman signing error on arch

joshrabinowitz:
For #916 

pacman was complaining
"libcap-2.65-1-x86_64.pkg.tar.zst is corrupted"
 See https://github.com/sobolevn/git-secret/runs/7564903813?check_suite_focus=true


joshrabinowitz:
I couldn't get this to work; I suspect it's a transient upstream packaging issue

PR #947, opened 2022-10-25
Ignore exact 

oliikit:

Currently, if a file is a top-level file within a repository and is added to git secret, the filename added to `.gitignore` will subsequently ignore all files with that name in further sub-directories.

This can cause some unintended behaviors if you are using git-secret in a mixed repository where there are both encrypted and unencrypted files as any file within sub-directories will be ignored. Adding files to both the `.gitignore` and `.git-secret/mapping.cfg` utilize [`git ls-files`](https://git-scm.com/docs/git-ls-files) to output the filename and then prepends the path to it in subsequent functions. To fix this issue, if a file is within the root directory (i.e., the path does not contain a `/`), I prepended a `/` to the filename so that git ignore works as intended. 

No

Steps to Reproduce
 _Current version: git-secret 0.5.0-alpha2_

1. Add a file within the root directory to git-secret
     1.  the file in both `.gitignore` and `git-secret` will be `FILENAME`
2. Commit the file
3. Add a file in a sub-directory with the _same_ name as in step 1. 
    1. _Result:_ file will be ignored - even if was intended to be added to git

Any other comments?
-------------------

While this can be fixed by editing the `.gitignore` whenever it causes conflict, this PR just stops the behavior from stopping it in the first place.

### DIFF
--- a/src/_utils/_git_secret_tools.sh
+++ b/src/_utils/_git_secret_tools.sh
@@ -304,6 +304,12 @@ function _git_normalize_filename {
 
   local result
   result=$(git ls-files --full-name -o "$filename")
+
+  # if file is in the top level, append a / so only that top level file is considered
+  if [[ $filename != *"/"* ]]; then
+    result="/$result"
+  fi
+
   echo "$result"
 }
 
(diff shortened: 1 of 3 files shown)


oliikit:
@joshrabinowitz - I've been meaning to check out the different platform errors. Admittedly, I'm on macOS 12.6 and had passing tests locally. I'm planning on taking a look at the other failed platform tests later this week.

Issue #981, opened 2023-02-26
Website not working


burhankhanzada:
This issue is not related to the actual git-secret but the website because I need to follow the documentation

![Screenshot 2023-02-26 175702](https://user-images.githubusercontent.com/18241005/221411808-b94cc0f6-82b2-4917-a196-ee33d535c3e2.png)

sobolevn:
I don't have any way to pay for the domain because of the sanctions.
We will switch to github-based domain.

zakupower:
For people who still need to access the website, while it is down. Wayback machine has a saved copy from end of last year.
https://web.archive.org/web/20221202055843/https://git-secret.io/git-secret-tell

joshrabinowitz:
@sobolevn any progress on this? Can I offer any assistance?

sobolevn:
Thanks for the notification! New domain: https://sobolevn.me/git-secret/

traviswimer:
@sobolevn I would consider this to be a significant security concern.

There are many links pointing to that expired domain and it is currently the top search result on Google. It's probably not good to risk the possibility of someone with malicious intent purchasing the domain of a widely used security tool.

Is it possible to transfer the domain to another registrar?

sobolevn:
No, they have completely banned my account. See https://github.com/sobolevn/git-secret/issues/790

There's nothing I can do :(

PR #982, opened 2023-03-18
Fix the domain name / path in the Jekyll config

sobolevn:
Closes https://github.com/sobolevn/git-secret/issues/981


sobolevn:
Let's try it!

Issue #786, opened 2022-02-27
automate packaging for arch linux


joshrabinowitz:
We have automated _testing_ on `arch`, now let's automate building `git-secret` package for `arch`.
See notes in #785

PR #1031, opened 2023-10-13
Automated packaging on arch linux

Night-Amber3301:
solves : #786

Automation of building git-secret package for arch.

…

…

(If it’s long, please paste to https://ghostbin.com/ and insert the link here.)

Any other comments?
-------------------
…


sobolevn (commented):
Thanks for the PR! But, I don't quite understand what it does:
1. It uses `docker-ci` image which we use only for testing, not for releasing
2. It uses a cleanup as a separate step, how effective is it?

Issue #84, opened 2017-03-05
Support `git diff` and `git log` via `.gitattributes`


sobolevn:
It is possible to configure standard `git` commands to use custom plugin's logic.
We need a note in the docs about that.

cupen:
A tool similar as `git secret view path/to/file` would be better.
`git secret cat` does'nt works for `path/to/file`

my git setup.
```
$ echo "*.secret diff=secret merge=secret" >  .gitattributes
$ git config diff.secret.textconv	     "git secret cat"
$ git config diff.secret.cachetextconv "false"
```

run `git diff`
```
$ git diff
gpg: can't open '/home/cupen/workbench/repos/badwords//tmp/FWIcAl___skip__.txt.secret': No such file or directory
gpg: decrypt_message failed: No such file or directory
git-secret: abort: problem decrypting file with gpg: exit code 2: /home/cupen/workbench/repos/badwords//tmp/FWIcAl___skip__.txt.secret
fatal: unable to read files to diff
```

jpbochi:
FWIW, transcrypt supports that. See https://github.com/elasticdog/transcrypt#designate-a-file-to-be-encrypted. I'm posting it here as a possible source for inspiration, not as a promotion of it.

icy:
> FWIW, transcrypt supports that. See https://github.com/elasticdog/transcrypt#designate-a-file-to-be-encrypted. I'm posting it here as a possible source for inspiration, not as a promotion of it.

There is also sops, though there is an issue https://github.com/mozilla/sops/issues/666 fyi

jpbochi:
I managed to make `git diff` work by calling `gpg` directly. This is what I did:

```
echo "*.secret diff=gpg" >  .gitattributes
git config --local diff.gpg.textconv 'gpg --decrypt 2> /dev/null'
```

PR #1428, opened 2026-08-27
docs: document git diff/log support via .gitattributes

TrueFurina:
## Summary

Per #84: standard `git` commands can use custom logic for secret files via `.gitattributes`. Adds a note in the man page documenting how.

## Changes

- `man/man7/git-secret.7.md`: new section **"Using git diff and git log with encrypted files"** covering:
  - mark secret files with a custom diff driver in `.gitattributes` (`*.secret diff=gpg`)
  - configure the `textconv` decrypt command (`git config --local diff.gpg.textconv 'gpg --decrypt 2> /dev/null'`)
  - the display-only nature (repo still stores ciphertext; viewers need the private key)

## Verification

- Pattern is the same approach reported working in the issue comments (gpg textconv)
- Markdown structure matches the surrounding man page sections


sobolevn (commented):
sorry, this seems strange. why all file is changed? did you run the `man` generation command? why did you change other parts of the file? can we keep this limited to the `.gitattributes`?

